PR #3, opened 2025-07-28
Updated to work with Python version 3.10 - 3.14

mcgicjn2:
Self explanatory, updated logic and readme for python 3.13 and 3.14 (future proof)


indigocoral (approved):
Seems good to go,

PR #16, opened 2025-08-18
Elementtest refactor to separate functionalities

indigocoral:
**Description:**
The current implementation mixes some of the UI logic (Tkinter components) with core logic (Kratos execution). This tight coupling makes the code harder to maintain, test, and extend further in the future.

**Goal:**
Refactor the codebase to establish a clear separation of the GUI and the core, where:

- Core logic (file preparation, Kratos execution, result parsing) is encapsulated in independent modules without UI dependencies.
- UI functionality (Tkinter, buttons, logging widgets) to be separated into different modules.

**Dependency**
Here is the dependency of the current workflow:
```mermaid
flowchart LR
  %% ===================== NODES =====================
  subgraph UI[UI layer]
    A[kratos_element_test_gui.py]
    B[ui_menu.py]
    C[ui_builder.py]
    D[controller.py]
    L[ui_labels.py]
    G[ui_logger.py]
    R[ui_runner.py]
  end

  subgraph P[Plotters]
    M[matplotlib_plotter.py]
    PL[plotter_labels.py]
  end

  subgraph Core
    subgraph IO[core/io]
      E1[mdpa_editor.py]
      E2[project_parameter_editor.py]
      E3[material_editor.py]
      E4[gid_output_reader.py]
      E5[udsm_parser.py]
    end
    subgraph Pipe[core/pipeline]
      RS[run_simulation.py]
      GTR[generic_test_runner.py]
    end
    CL[core_labels.py]
    IF[interfaces.py]
    MDL[models.py]
  end

  %% ===================== IMPORT / ORCHESTRATION (SOLID) =====================
  %% UI boot & wiring
  A --> B --> C --> D
  C --> G
  D --> G
  R --> RS

  %% UI now imports the parser from Core/IO
  B --> E5

  %% Core uses IO editors/readers
  RS --> E1
  RS --> E2
  RS --> E3
  RS --> E4

  %% Plotting is injected/used by UI (no core import of plotters)
  C --> M
  D --> M
  M --> PL

  %% Type contracts / models (module-level relationships)
  RS --> IF

  %% ===================== RUNTIME FLOW / DATA (DOTTED) =====================
  %% Controller passes user selections & inputs into the core
  D -. "test_type, drainage,\nMC indices, parameters" .-> RS

  %% Core returns arrays/results that the plotter renders
  RS -. "results arrays" .-> M

  %% Plotter + Controller log to UI logger
  M -. "warnings/info" .-> G
  D -. "status/progress" .-> G

  %% Controller builds/validates input dataclasses
  D -. "builds SimulationInputs" .-> MDL

  %% Plotter conforms to the Plotter Protocol
  M -. "conforms to Plotter (Protocol)" .-> IF

  %% ===================== LEGEND =====================
  %% Solid arrow: import / call orchestration
  %% Dotted arrow: runtime data/results or type contract (no hard dependency)
```

### DIFF
--- a/kratos_element_test/ui/ui_utils.py
+++ b/kratos_element_test/ui/ui_utils.py
@@ -15,7 +15,9 @@ def _asset_path(name: str) -> str:
        Falls back to <this_dir>/assets/<name> if not found.
        """
     ui_dir = Path(__file__).resolve().parent
-    for p in (ui_dir / "assets" / name, ui_dir.parent / "assets" / name):
+    candidates = [ui_dir / "assets" / name, ui_dir.parent / "assets" / name]
+    for p in candidates:
         if p.exists():
             return str(p)
-    return str(ui_dir / "assets" / name)
+    raise FileNotFoundError(
+        f"Asset '{name}' not found. Tried: {', '.join(str(p) for p in candidates)}")


kratos_element_test/ui/ui_runner.py
```diff
@@ -3,10 +3,18 @@
...
-from kratos_element_test.run_simulation import run_simulation
+import warnings
+from kratos_element_test.ui.ui_logger import log_message
+from kratos_element_test.core.pipeline.run_simulation import run_simulation
+from kratos_element_test.plotters.matplotlib_plotter import MatplotlibPlotter
+
 
 def run_gui_builder(test_type, dll_path, index, material_parameters, input_widgets, cohesion_phi_indices, axes):
+
+    warnings.warn("ui_runner.run_gui_builder is deprecated. UI now uses ElementTestController "
```
rfaasse, line 14: I don't see the run_gui_builder function is used anywhere, can't we just remove this file, instead of warning for deprecation?

kratos_element_test/ui/ui_menu.py
```diff
@@ -7,10 +7,10 @@
 from tkinter import filedialog, messagebox, ttk, scrolledtext, Menu
 from platformdirs import user_data_dir
 from pathlib import Path
-from kratos_element_test.ui_builder import GeotechTestUI
-from kratos_element_test.ui_udsm_parser import udsm_parser
-from kratos_element_test.ui_labels import (APP_TITLE, APP_VERSION, APP_NAME, APP_AUTHOR, SELECT_UDSM, LINEAR_ELASTIC,
-                                           HELP_MENU_FONT, DEFAULT_TKINTER_DPI)
+from kratos_element_test.ui.ui_builder import GeotechTestUI
+from kratos_element_test.core.io.udsm_parser import udsm_parser
```
rfaasse, line 11: Nice that this was moved, let's see if in next PRs we can make sure it's also only used in core/model and controllers

kratos_element_test/ui/ui_builder.py
```diff
@@ -292,37 +352,34 @@ def _run_simulation(self):
...
-                )
+                w = self.shear_widgets
+                tt = "direct_shear"
+
+            sigma_init = float(w["Initial effective cell pressure |σ'ₓₓ|"].get())
+            eps_max = float(w["Maximum Strain |εᵧᵧ|"].get())
+            n_steps = float(w["Number of steps"].get())
+            duration = float(w["Duration"].get())
+
+            self.controller.run(
```
rfaasse, line 373: Ideally (again for later, just writing down my thoughts), the ui it doesn't call the controller directly, but emits some kind of signal the controller can pick up (same holds for the other calls). However I don't really know how this works in tkinter, or if it's even possible.

kratos_element_test/ui/ui_builder.py
```diff
@@ -175,6 +201,7 @@ def _create_input_fields(self):
 
         self._switch_test(TRIAXIAL)
 
+        # self._init_drainage_section()
```
rfaasse, line 204: If we don't need this code anymore, could you please remove it?

kratos_element_test/ui/controller.py
rfaasse, line 1: I think it would be a good idea to have the name of the file the same as the name of the class (in this case element_test_controller.py)

rfaasse (commented):
Thanks for the effort in changing the structure of this piece of software for the better! I think the file structure is a lot clearer, the templates, assets, ui, plotter and core components are where I would expect them to be! 

Most of my comments are minor and/or questions. I also added a few comments to illustrate a bit more concretely what the next steps are for next PRs. Basically the location of functionality has improved a lot, it's now a matter of improving 'who calls who' and who keeps the state.

A more generic question: how do we make sure all functionality still works. I didn't test the tool again yet, but should we think of a structured approach to make sure everything still works as expected after these kinds of changes?

kratos_element_test/ui/ui_builder.py
```diff
@@ -2,27 +2,37 @@
...
     INPUT_SECTION_FONT, HELP_MENU_FONT
 )
 
+
+def _asset_path(name: str) -> str:
+    ui_dir = Path(__file__).resolve().parent
+    for p in (ui_dir / "assets" / name, ui_dir.parent / "assets" / name):
+        if p.exists():
+            return str(p)
+    return str(ui_dir / "assets" / name)
```
avdg81, line 33: My understanding is that this function returns the path to the assets directory. To that end, it considers two locations and it will return the first one that exists. If, however, neither of them exists, we return the path to the first location for which we already know that it doesn't exist. This implies that any clients calling this function must anticipate the possibility that the path doesn't exist. My question is why? As far as I understand, this isn't user input, but something that is fully under our control. Shouldn't we simply assert that the file path must be there, or else it's considered an error?

kratos_element_test/ui/ui_builder.py
```diff
@@ -140,8 +167,8 @@ def _create_input_fields(self):
         self.test_buttons = {}
 
         image_paths = {
-            TRIAXIAL: os.path.join(os.path.dirname(__file__), "assets", "Triaxial.png"),
-            DIRECT_SHEAR: os.path.join(os.path.dirname(__file__), "assets", "DSS.png")
+            TRIAXIAL:  _asset_path("Triaxial.png"),
+            DIRECT_SHEAR: _asset_path("DSS.png")
```
avdg81, line 171: Much clearer. Thanks :-)

kratos_element_test/ui/ui_builder.py
```diff
@@ -292,37 +352,34 @@ def _run_simulation(self):
...
-                    test_type="direct_shear",
-                    dll_path=self.dll_path or "",
-                    index=index,
-                    material_parameters=[float(x) for x in material_params],
-                    input_widgets=self.shear_widgets,
-                    cohesion_phi_indices=cohesion_phi_indices,
-                    axes=self.axes
-                )
+                w = self.shear_widgets
+                tt = "direct_shear"
```
avdg81, line 366: I was wondering whether we could implement a mapping from the test type (i.e. `TRIAXIAL` and `DIRECT_SHEAR`) to the corresponding test type names (`"triaxial"` and `"direct_shear"`)? In that way, we need to define it just once. As I commented before, this mapping is now repeated a few times.

kratos_element_test/ui/ui_builder.py
avdg81, line 1: This Python script contains lots of GUI-specific details that I'm not familiar with. I've tried to review this file to the best of my abilities, but my feedback here might be insufficient.

kratos_element_test/ui/ui_labels.py
avdg81, line 1: Since this file doesn't contain labels only (but also font names and a default resolution), we may want to give it a more general name. What about `ui_constants.py`?

kratos_element_test/ui/ui_menu.py
```diff
@@ -21,9 +21,15 @@
...
 
 
-def show_license_agreement(readonly=False):
-    license_file_path = os.path.join(os.path.dirname(__file__), "assets", "license.txt")
+def _asset_path(name: str) -> str:
+    ui_dir = Path(__file__).resolve().parent
+    for p in (ui_dir / "assets" / name, ui_dir.parent / "assets" / name):
+        if p.exists():
+            return str(p)
+    return str((ui_dir / "assets" / name))
```
avdg81, line 29: I think this is a duplicate of the same-named function in `ui_builder.py`. I'd like to suggest to extract it to a utilities file and call it from both usages.

kratos_element_test/ui/ui_runner.py
avdg81, line 1: Since the only function contained in this file isn't called from anywhere in your code, I'd suggest to remove the file completely.

avdg81 (commented):
Hi Marjan,

Thanks so much for separating the core functionality from the GUI functionality. I feel that the code has become a lot more modular, and therefore easier to understand, maintain, and extend.

I have several suggestions and questions, but none of them is blocking. All in all, I feel this is a major improvement, so thanks for all the hard work you've done.

kratos_element_test/core/io/mdpa_editor.py
```diff
@@ -31,21 +34,23 @@ def update_maximum_strain(self, maximum_strain):
         replacer = self._replacer_factory(prescribed_displacement)
         new_text, count = re.subn(pattern, replacer, self.raw_text)
         if count == 0:
-            log_message("Could not update maximum strain.", "warn")
+            self._log("Could not update maximum strain.", "warn")
         else:
             self.raw_text = new_text
             MdpaEditor.save(self)
+            self.save()
```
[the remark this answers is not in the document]
indigocoral: Good catch! Removed.

kratos_element_test/core/templates/__init__.py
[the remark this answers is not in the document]
indigocoral: I tried to explain it a bit in your previous comment. see here: https://github.com/Deltares-research/KratosElementTestSuite/pull/16#discussion_r2287490254
But maybe it's best if we discuss whether to keep them or whether they can be removed.

kratos_element_test/ui/ui_menu.py
indigocoral: Check!

kratos_element_test/ui/controller.py
indigocoral: Good point! Done.

kratos_element_test/core/templates/__init__.py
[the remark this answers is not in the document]
indigocoral: All removed.

kratos_element_test/core/io/project_parameter_editor.py
```diff
@@ -4,20 +4,23 @@
 
 import re
 import json
-from tkinter import messagebox
 
 
+def _fallback_log(msg: str, level: str = "info"):
+    print(f"{level.upper()}: {msg}")
```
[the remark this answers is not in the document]
indigocoral: Done.

kratos_element_test/ui/ui_runner.py
indigocoral: Yes, it has been removed.

kratos_element_test/core/io/project_parameter_editor.py
```diff
@@ -4,20 +4,23 @@
...
+        self._log = logger or _fallback_log
         with open(self.json_path, 'r') as f:
             self.raw_text = f.read()
 
     def _write_back(self):
         with open(self.json_path, 'w') as f:
             f.write(self.raw_text)
 
-    def update_nested_value(self, module_name, key, new_list):
+    def update_nested_value(self, module_name, key, new_list) -> bool:
```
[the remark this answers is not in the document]
indigocoral: Yes! Good catch! Initially, I had the idea to pass this boolean value to the logger: True on success and False when nothing was updated, while also logging a warning. But then I changed my mind but forgot to remove the return. It has been fixed now! Thanks!

kratos_element_test/core/io/project_parameter_editor.py
```diff
@@ -29,20 +32,22 @@ def update_nested_value(self, module_name, key, new_list):
...
 
             self.raw_text = json.dumps(data, indent=4)
             self._write_back()
+            return True
         except Exception as e:
-            messagebox.showerror("Error", f"Failed to update stress vector: {e}")
+            raise RuntimeError(f"Failed to update '{key}' under '{module_name}': {e}") from e
 
-    def update_property(self, property_name, new_value):
+    def update_property(self, property_name, new_value) -> int:
```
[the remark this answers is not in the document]
indigocoral: Yes this is same as the boolean for `update_nested_value`:
Initially, I had the idea to pass this `int` as a number of updated values to the logger. But then I changed my mind but forgot to remove the return. Thanks!

kratos_element_test/core/interfaces.py
```diff
@@ -0,0 +1,21 @@
...
+                 eps_yy: Iterable[float],
+                 diff_sig: Iterable[float],
+                 sigma1: Iterable[float],
+                 sigma3: Iterable[float],
+                 mean_stress: Iterable[float],
+                 von_mises: Iterable[float],
+                 cohesion: Optional[float],
+                 phi: Optional[float]) -> None: ...
+
+    def direct_shear(self, *args, **kwargs) -> None: ...
```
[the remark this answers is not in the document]
indigocoral: Good catch indeed! Yes, this was left unintentionally. I have removed it now. Thank you.

kratos_element_test/ui/controller.py
```diff
@@ -0,0 +1,109 @@
...
+            index: Optional[int],
+            material_parameters: List[float],
+            sigma_init: float,
+            eps_max: float,
+            n_steps: float,
+            duration: float) -> None:
+
+
+        tt = test_type or self._test_type
+        if tt not in ("triaxial", "direct_shear"):
```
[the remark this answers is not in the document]
indigocoral: Good idea. Done

kratos_element_test/ui/ui_builder.py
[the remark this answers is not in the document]
indigocoral: Removed.

kratos_element_test/ui/ui_builder.py
```diff
@@ -217,19 +244,45 @@ def _create_mohr_options(self, params):
...
             log_message("Mohr-Coulomb model is selected.", "info")
+
+            self.controller.set_mohr_enabled(True)
+
+            try:
+                c_idx = int(self.cohesion_var.get()) if self.cohesion_var.get() else None
+                phi_idx = int(self.phi_var.get()) if self.phi_var.get() else None
+
+            except ValueError:
+                c_idx, phi_idx = None, None
```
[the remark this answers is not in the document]
indigocoral: Very good idea! Done.

kratos_element_test/ui/ui_builder.py
```diff
@@ -217,19 +244,45 @@ def _create_mohr_options(self, params):
...
+            self.controller.set_mohr_enabled(False)
+            self.controller.set_mohr_mapping(None, None)
+
             for w in widgets:
                 w.pack_forget()
 
     def _switch_test(self, test_name):
         clear_log()
         self.current_test.set(test_name)
+        self.controller.set_test_type("triaxial" if test_name == TRIAXIAL else "direct_shear")
```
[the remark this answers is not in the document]
indigocoral: Furthermore, I added valid test types and valid drainage types to the ui_constants.py to hold these constants in one place. Thus I did a bit of clean up for that.

kratos_element_test/ui/ui_builder.py
indigocoral: Done

kratos_element_test/plotters/matplotlib_plotter.py
```diff
@@ -3,16 +3,57 @@
...
-    TITLE_DIFF_PRINCIPAL_SIGMA_VS_STRAIN, TITLE_VOL_VS_VERT_STRAIN, TITLE_MOHR, TITLE_P_VS_Q, TITLE_SHEAR_VS_STRAIN,
+    MOBILIZED_SHEAR_STRESS_LABEL, P_STRESS_LABEL, Q_STRESS_LABEL,
+    TITLE_SIGMA1_VS_SIGMA3, TITLE_DIFF_PRINCIPAL_SIGMA_VS_STRAIN, TITLE_VOL_VS_VERT_STRAIN,
+    TITLE_MOHR, TITLE_P_VS_Q, TITLE_SHEAR_VS_STRAIN,
     LEGEND_MC, LEGEND_MC_FAILURE
 )
 
 
+def log_message(msg: str, level: str = "info"):
+    print(f"{level.upper()}: {msg}")
```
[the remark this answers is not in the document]
indigocoral: Done

rfaasse (commented):
Thanks for processing the review comments! I have a few very minor suggestions/questions left after the latest changes, but this is close to merging!

kratos_element_test/core/io/project_parameter_editor.py
```diff
@@ -23,26 +24,29 @@ def update_nested_value(self, module_name, key, new_list):
...
         replacement = rf'\g<1>{new_value}'
         self.raw_text, count = re.subn(pattern, replacement, self.raw_text)
         if count == 0:
-            messagebox.showwarning("Warning", f"Could not find '{property_name}' to update.")
+            self._log(f"Could not find '{property_name}' to update.", "warn")
         elif count > 1:
-            messagebox.showwarning("Warning", f"Multiple occurrences of '{property_name}' found. Updated all {count}.")
+            self._log(f"Multiple occurrences of '{property_name}' found. Updated all {count}.", "warn")
         self._write_back()
+        return None
```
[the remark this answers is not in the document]
indigocoral: Thanks for spotting this. You’re right, the mixed return values (False vs None) were confusing. I’ve changed both update_nested_value and update_property to behave consistently as void methods (they always return None), with outcome reporting handled via logging or exceptions.

This way both methods are consistent:

- Success or no-op → None
- Warnings → logged
- Errors → exceptions

Callers don’t need to check return values and can rely on logging instead.

kratos_element_test/plotters/matplotlib_plotter.py
```diff
@@ -13,14 +14,10 @@
...
 
-def log_message(msg: str, level: str = "info"):
-    print(f"{level.upper()}: {msg}")
-
 class MatplotlibPlotter:
     def __init__(self, axes, logger=None):
         if logger is not None:
-            global log_message
-            log_message = logger
+            self._log = logger or _fallback_log
```
avdg81, line 20: Note that in this way attribute `_log` is only defined when one is supplied (i.e. when `logger` is not `None`). I'm not sure whether that is what you intended? Wouldn't it be better to always define attribute `_log` and set it to the fallback logger when no logger is provided. In other words:
```python
def __init__(self, axes, logger=None):
    self._log = logger or _fallback_log
    self.axes = list(axes or [])
```

kratos_element_test/ui/element_test_controller.py
```diff
@@ -35,12 +42,12 @@ def _mc_tuple(self) -> Optional[Tuple[int, int]]:
...
-        if test_type not in ("triaxial", "direct_shear"):
-            self._logger(f"Unknown test type: {test_type}", "warn")
+        if not self._is_valid_test_type(test_type):
+            return
         self._test_type = test_type
 
     def set_drainage(self, drainage: str) -> None:
-        if drainage not in ("drained", "undrained"):
+        if drainage not in VALID_DRAINAGE_TYPES:
             self._logger(f"Unknown drainage: {drainage}", "warn")
```
avdg81, line 51: Question: wouldn't you want to return when the drainage type is found to be invalid? Such behavior would be consistent with how you handle invalid test types.

kratos_element_test/core/io/project_parameter_editor.py
```diff
@@ -50,4 +60,4 @@ def update_property(self, property_name, new_value) -> int:
         elif count > 1:
             self._log(f"Multiple occurrences of '{property_name}' found. Updated all {count}.", "warn")
         self._write_back()
-        return count
+        return None
```
avdg81, line 63: I think this `return` statement is redundant. Leaving it out has the same effect.

kratos_element_test/core/io/project_parameter_editor.py
```diff
@@ -20,28 +30,28 @@ def _write_back(self):
...
-
-            self.raw_text = json.dumps(data, indent=4)
-            self._write_back()
-            return True
+                    self.raw_text = json.dumps(data, indent=4)
+                    self._write_back()
+                    return
+
+            self._log(f"Could not find '{key}' under '{module_name}'.", "warn")
+            return
```
avdg81, line 49: I think this `return` statement is redundant. There is no code that follows it, so an implicit return will do.

avdg81 (commented):
Thanks so much for processing the review suggestions. I feel we're almost ready to merge this PR. I have a few tiny suggestions that you may want to apply in addition to the changes you've already made. Hope this helps.

kratos_element_test/core/io/material_editor.py
```diff
@@ -3,12 +3,14 @@
 # Contact kratos@deltares.nl
 
 import json
+from kratos_element_test.core.core_utils import _fallback_log
 
 
 class MaterialEditor:
-    def __init__(self, json_path):
+    def __init__(self, json_path, logger=None):
```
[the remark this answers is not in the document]
rfaasse: Nice and clear, thanks!

rfaasse (approved):
Thanks for processing the review comments, looks good to go to me!

avdg81 (approved):
Thanks so much for taking care of the final review suggestions. I feel this PR can be merged now.

PR #31, opened 2025-12-10
Separate running and obtaining results

rfaasse:
This PR separates functionality for running a simulation and collecting its results by:
- Creating a new class for collecting the data: `ResultCollector`
- Moving functionality from the `GenericTestRunner` and `RunSimulation` classes related to collecting results to the `ResultCollector`
- Make two separate calls in the `RunSimulation`, such that the controller can call them separately
- Moved the render functionality to the controller, such that the model/core doesn't have to know about plotting (small bonus)

### DIFF
--- a/kratos_element_test/model/test/test_result_collector.py
+++ b/kratos_element_test/model/test/test_result_collector.py
@@ -1,17 +1,19 @@
 import os
 import unittest
 from pathlib import Path
+
 import numpy as np
 from parameterized import parameterized
 
-from kratos_element_test.model.core_utils import (
-    seconds_to_hours,
-    second_list_to_hour_list,
-)
+from kratos_element_test.model.core_utils import seconds_list_to_hours_list
 from kratos_element_test.model.pipeline.result_collector import ResultCollector
 
 
 class ResultCollectorTest(unittest.TestCase):
+
+    def setUp(self):
+        self.test_path = Path(os.path.dirname(__file__)) / "output.post.res"
+
     @parameterized.expand(
         [
             ["yy_strain", [0.0, -0.05, -0.1, -0.15, -0.2]],
@@ -28,15 +30,14 @@ class ResultCollectorTest(unittest.TestCase):
             ["sigma_yy", [-100.0, -45100.0, -90100, -135100, -180100]],
             [
                 "time_steps",
-                second_list_to_hour_list([0.2, 0.4, 0.6, 0.8, 1.0]),
+                seconds_list_to_hours_list([0.2, 0.4, 0.6, 0.8, 1.0]),
             ],
         ]
     )
     def test_collected_values(self, variable_name, expected_values):
-        file_path = Path(os.path.dirname(__file__))
-        test_path = file_path / "output.post.res"
+
         collector = ResultCollector(
-            [test_path],
+            [self.test_path],
             material_parameters=[1.0, 2.0, 3.0, 4.0],
             cohesion_phi_indices=(3, 4),
         )
@@ -56,10 +57,8 @@ class ResultCollectorTest(unittest.TestCase):
             self.fail(f"collect_results raised an exception unexpectedly: {e}")
 
     def test_c_phi_are_none_when_indices_are_not_defined(self):
-        file_path = Path(os.path.dirname(__file__))
-        test_path = file_path / "output.post.res"
         collector = ResultCollector(
-            [test_path],
+            [self.test_path],
             material_parameters=[1.0, 2.0, 3.0, 4.0],
             cohesion_phi_indices=(),
         )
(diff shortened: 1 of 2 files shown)


kratos_element_test/core/pipeline/run_simulation.py
```diff
@@ -81,51 +67,38 @@ def __init__(
...
-                "mean_stress": mean_stress,
-                "von_mises": von_mises,
-                "cohesion": cohesion,
-                "phi": phi,
-                "sigma_xx": sigma_xx,
-                "sigma_yy": sigma_yy,
-                "time_steps": time_steps,
-            }
-
-            self._render(results)
```
rfaasse: This line I moved to the controller, such that the model doesn't have to care about displaying the results

kratos_element_test/core/pipeline/run_simulation.py
```diff
@@ -81,51 +67,38 @@ def __init__(
...
+            )
+            results = collector.collect_results()
             self.log("Rendering complete.", "info")
             return results
 
         finally:
+            # TODO check whether the cleanup is working correctly now
             if self.keep_tmp:
                 print(f"[Info] Temporary folder retained at: {self.tmp_dir}")
             else:
```
rfaasse: This we need to discuss a bit: Right now I can only remove the tmp folder after post processing. In theory, we could run multiple scenarios and only post process it once (so tmp folders wouldn't be cleaned up). Right now the controller always calls both methods, so I expect the functionality of clean-up to be the same, but I think it's good to double check this

kratos_element_test/ui/element_test_controller.py
```diff
@@ -114,17 +114,64 @@ def run(self,
...
+                results["sigma_xx"],
+                results["mean_stress"],
+                results["von_mises"],
+                results["sigma1"],
+                results["sigma3"],
+                results["cohesion"],
+                results["phi"],
+            )
+        else:
+            raise ValueError(f"Unsupported test_type: {test_type}")
```
rfaasse: As you can see I moved the responsibility of plotting to the controller, such that the model/core doesn't have to know about plotting. Core just runs and retrieves the results. The GUI will handle how the results are displayed. 

Nice thing is that you already made the model dependent only on this interface of the plotter, not the details, so that made moving this around very easy 👍

kratos_element_test/core/test/test_result_collector.py
indigocoral: Thanks very much for adding these tests! They will help a lot in the future developments.

kratos_element_test/core/test/test_result_collector.py
```diff
@@ -0,0 +1,110 @@
...
+        test_path = file_path / "output.post.res"
+        collector = ResultCollector(
+            [test_path, test_path],
+            material_parameters=[1.0, 2.0, 3.0, 4.0],
+            cohesion_phi_indices=(3, 4),
+        )
+        results = collector.collect_results()
+
+        # Since the same results file is used twice, the stress values repeat
+        # the same values for the second stage
```
avdg81: This comment (and the one at line 93) is really helpful. Thanks.

indigocoral (approved):
Richard, thank you very much for this refactoring. It was very easy to follow the changes. They all made sense to me and I did not see anything that requires changing same as the previous one. Only praise comments :)
I appreciate how you have kept the code as is.

kratos_element_test/model/test/test_result_collector.py
```diff
@@ -0,0 +1,110 @@
...
+            ["phi", [4.0]],
+            ["sigma_xx", [-100.0, -100.0, -100.0, -100.0, -100.0]],
+            ["sigma_yy", [-100.0, -45100.0, -90100, -135100, -180100]],
+            [
+                "time_steps",  # in hours, hence the division by 3600
+                [0.2 / 3600, 0.4 / 3600, 0.6 / 3600, 0.8 / 3600, 1.0 / 3600],
+            ],
+        ]
+    )
+    def test_collected_values(self, variable_name, expected_values):
```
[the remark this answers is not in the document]
avdg81: Ah, now I see it. Thanks for explaining Richard. I think I've missed it, since `@parameterized.expand` is twenty lines above the test case name (due to inlining the provided list).

avdg81 (approved):
Hi Richard,
Thanks a lot for taking care of the review suggestions. I have two minor remarks, but there is no need to take action if you disagree with them. Feel free to merge this PR.

kratos_element_test/core/pipeline/result_collector.py
```diff
@@ -0,0 +1,246 @@
...
+            yy_strain_stages.append(yy_strain)
+            all_vol_strain.extend(vol_strain)
+            all_shear_strain_xy.extend(shear_strain_xy)
+            all_von_mises.extend(von_mises_values)
+            all_mean_stress.extend(mean_stress_values)
+            all_sigma_xx.extend(sigma_xx)
+            all_sigma_yy.extend(sigma_yy)
+            all_time_steps.extend(
+                [t / 3600.0 for t in time_steps]
+            )  # Convert seconds → hours
```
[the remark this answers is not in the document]
rfaasse: That's also a good point. Since this is a GUI specifically for Kratos anyway, I think it isn't that big of a problem to just use the kratos functionality in the 'model' part here. In a next PR I'll have a look how easy it is to access (also in a GitHub pipeline)

PR #38, opened 2025-12-22
Integrate internal mohr coulomb into the suite

indigocoral:
This PR is to add the internal Mohr-Coulomb model within Kratos to the ElementTestSuite alongside Linear Elastic model and the ability to choose a soil DLL. The option of plotting the Mohr failure envelope is made to be automatic with the selection of Mohr-Coulomb model.

### DIFF
--- a/kratos_element_test/model/pipeline/run_simulation.py
+++ b/kratos_element_test/model/pipeline/run_simulation.py
@@ -215,7 +215,6 @@ class RunSimulation:
                     "GEO_DILATANCY_ANGLE": self.material_parameters[5],
                 })
                 editor.set_constitutive_law("GeoMohrCoulombWithTensionCutOff2D")
-                return
 
     def _set_project_parameters(self) -> None:
         with open(self.project_json_path, "r") as f:


kratos_element_test/view/ui_builder.py
```diff
@@ -262,6 +258,30 @@ def _toggle_mohr_options(self):
...
+
+        elif self.is_mohr_coulomb:
+            self.controller.set_mohr_enabled(True)
+            self.cohesion_var = tk.StringVar(value="3")
+            self.phi_var = tk.StringVar(value="4")
+
+            c_idx, phi_idx = self._parse_mc_indices()
+            self.controller.set_mohr_mapping(c_idx, phi_idx)
+
+            self.mohr_checkbox_widget.configure(state="disabled")
```
rfaasse: To me as a user it's a bit counter-intuitive that the checkbox is disabled and not checked when I have a Mohr Coulomb model selected (though I understand as a developer why it's necessary). Would it be possible to hide it instead of disabling (same for the linear elastic)?

kratos_element_test/view/ui_menu.py
```diff
@@ -199,18 +199,45 @@ def load_linear_elastic():
...
+                "param_names": [[
+                    "Young's Modulus",
+                    "Poisson's Ratio",
+                    "Cohesion",
+                    "Friction Angle",
+                    "Tensile Strength",
+                    "Dilatancy Angle",
+                ]],
+                "param_units": [["kN/m²", "-", "kN/m²", "deg", "kN/m²", "deg"]],
+            }
```
rfaasse: For now it's fine to create the model data here and store it in the GUI, later we'll save it in the 'model' part of the application 👍

kratos_element_test/view/ui_builder.py
```diff
@@ -262,6 +258,30 @@ def _toggle_mohr_options(self):
...
+            self.controller.set_mohr_enabled(False)
+            self.controller.set_mohr_mapping(None, None)
+            self.mohr_checkbox_widget.configure(state="disabled")
+
+        elif self.is_mohr_coulomb:
+            self.controller.set_mohr_enabled(True)
+            self.cohesion_var = tk.StringVar(value="3")
+            self.phi_var = tk.StringVar(value="4")
+
+            c_idx, phi_idx = self._parse_mc_indices()
```
rfaasse: Do we need to specifically set the cohesion_var and phi_var, or is it only used for finding the c_idx and phi_idx? In that case we could just write:

```suggestion
            c_idx = 3
            phi_idx = 4
```

indigocoral: I removed the duplicated lines, and hopefully this is ok now?

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -193,14 +195,27 @@ def _set_material_constitutive_law(self) -> None:
...
-                }
-            )
-            editor.set_constitutive_law("GeoLinearElasticPlaneStrain2DLaw")
+                    "GEO_COHESION": self.material_parameters[2],
+                    "GEO_FRICTION_ANGLE": self.material_parameters[3],
+                    "GEO_TENSILE_STRENGTH": self.material_parameters[4],
+                    "GEO_DILATANCY_ANGLE": self.material_parameters[5],
+                })
+                editor.set_constitutive_law("GeoMohrCoulombWithTensionCutOff2D")
+                return
```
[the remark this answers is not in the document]
indigocoral: Correct! Done.

rfaasse (approved):
Thanks for processing the comments! The 'hidden checkbox for in-built models' works as expected, so this addition can be merged!

PR #43, opened 2026-01-02
Finalize test input mvc workflow

rfaasse:
In this pull request, the model view control workflow for the test inputs is finalized. In short, this means that the data for the test inputs is stored in the model class and also retrieved from there when running the simulation.

Changelog:
- Created a new class for the SoilTestInputView
- Created a run function in the main model, which retrieves the relevant test input data from the test input manager instead of getting the inputs via arguments
- Added a validation function for the SoilTestInputView, to check that every entry is at least convertible to a number

### DIFF
--- a/kratos_element_test/controller/soil_test_input_controller.py
+++ b/kratos_element_test/controller/soil_test_input_controller.py
@@ -1,4 +1,7 @@
-from kratos_element_test.model.models import TriaxialAndShearSimulationInputs
+from kratos_element_test.model.models import (
+    TriaxialAndShearSimulationInputs,
+    CRSSimulationInputs,
+)
 from kratos_element_test.model.soil_test_input_manager import SoilTestInputManager
 from kratos_element_test.view.ui_constants import (
     TRIAXIAL,
@@ -26,7 +29,9 @@ class SoilTestInputController:
     def get_crs_inputs(self):
         return self._soil_test_input_manager.input_data.get(CRS)
 
-    def get_current_test_inputs(self):
+    def get_current_test_inputs(
+        self,
+    ) -> TriaxialAndShearSimulationInputs | CRSSimulationInputs:
         return self._soil_test_input_manager.get_current_test_inputs()
 
     def add_crs_strain_increment(self) -> None:


kratos_element_test/model/main_model.py
```diff
@@ -1,6 +1,48 @@
...
+            udsm_number=udsm_number,
+            material_parameters=material_parameters,
+            cohesion_phi_indices=mohr_coulomb_options.to_indices(),
+            logger=self._logger,
+        )
+
+        sim.run()
+        self._latest_results = sim.post_process_results()
+
+    def get_latest_results(self):
```
rfaasse: The nice part about the model saving the latest results, is that we could start retrieving the results from here instead of using the registered global ui instance that's there now!

kratos_element_test/controller/element_test_controller.py
```diff
@@ -54,101 +62,49 @@ def _is_valid_test_type(self, test_type: Optional[str]) -> bool:
...
+        self,
+        *,
+        axes,
+        model_name: Optional[str] = None,
+        dll_path: str,
+        udsm_number: Optional[int],
+        material_parameters: List[float],
+    ) -> bool:
+        test_type = TEST_NAME_TO_TYPE.get(
+            self._main_model.soil_test_input_manager.get_current_test_type()
```
avdg81: I was just wondering how accurate the name `get_current_test_type` is when we use it to look up the test type given test name (at least, that's what `TEST_NAME_TO_TYPE` suggests to me). Don't we get the test name from the soil test input manager?

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -31,38 +35,47 @@ class RunSimulation:
...
+
+        if isinstance(test_inputs, CRSSimulationInputs):
+            self.stage_durations = [
+                hours_to_seconds(inc.duration_in_hours)
+                for inc in test_inputs.strain_increments
+            ]
+            self.step_counts = [inc.steps for inc in test_inputs.strain_increments]
+            self.strain_incs = [
+                inc.strain_increment for inc in test_inputs.strain_increments
+            ]
```
avdg81: I'm doubting here a bit whether it wouldn't be clearer to have single assignments rather than setting a default value and then overwriting it when needed (just like you did for the initialization of `self.drainage`). Perhaps something along the following lines:
```suggestion
        is_crs_test = isinstance(test_inputs, CRSSimulationInputs)
        self.stage_durations = [
            hours_to_seconds(inc.duration_in_hours)
            for inc in test_inputs.strain_increments
        ] if is_crs_test else None
        self.step_counts = [inc.steps for inc in test_inputs.strain_increments] if is_crs_test else None
        self.strain_incs = [
            inc.strain_increment for inc in test_inputs.strain_increments
        ] if is_crs_test else None
```
Or perhaps use `self.test_type` to initialize `is_crs_test`. If you (or @indigocoral ) feel(s) this suggestion makes it less readable, feel free to skip this suggestion.

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -206,14 +219,16 @@ def _set_material_constitutive_law(self) -> None:
...
             if self.model_name.strip().lower() == "mohr-coulomb model":
-                editor.update_material_properties({
-                    "YOUNG_MODULUS": self.material_parameters[0],
-                    "POISSON_RATIO": self.material_parameters[1],
-                    "GEO_COHESION": self.material_parameters[2],
-                    "GEO_FRICTION_ANGLE": self.material_parameters[3],
-                    "GEO_TENSILE_STRENGTH": self.material_parameters[4],
-                    "GEO_DILATANCY_ANGLE": self.material_parameters[5],
-                })
+                editor.update_material_properties(
```
avdg81: Just something to think about: at present, `self.material_parameters` is a list of values (probably mostly numbers), which requires knowledge about which entry represents what to use it. Would it be possible in the future to let `self.material_parameters` be a map rather than a list of values, such that the key describes what the value represents (in much the same you build up a map here)?

kratos_element_test/model/soil_test_input_manager.py
```diff
@@ -68,10 +69,10 @@ def update_num_steps(self, new_steps: int, test_type: str) -> None:
         self.input_data[test_type].number_of_steps = new_steps
 
     def update_duration(self, new_duration: float, test_type: str) -> None:
-        self.input_data[test_type].duration = new_duration
+        self.input_data[test_type].duration_in_seconds = new_duration
```
avdg81: You could opt to make the name of the function parameter state its unit, too:
```suggestion
        self.input_data[test_type].duration_in_seconds = new_duration_in_seconds
```

kratos_element_test/model/soil_test_input_manager.py
```diff
@@ -68,10 +69,10 @@ def update_num_steps(self, new_steps: int, test_type: str) -> None:
...
 
-    def update_drainage(self, new_drainage: str, test_type: str) -> None:
-        self.input_data[test_type].drainage = new_drainage
+    def update_drainage(self, new_drainage: str) -> None:
+        self.get_current_test_inputs().drainage = new_drainage
```
avdg81: Nice cleanup 👍

kratos_element_test/model/test/test_soil_test_input_manager.py
```diff
@@ -101,6 +101,20 @@ def test_update_init_pressure(self, test_type):
         ].initial_effective_cell_pressure
         self.assertEqual(updated, 250.0)
 
+    def test_get_current_default_inputs(self):
+        inputs = self.input_manager.get_current_test_inputs()
+        self.assertIsNotNone(inputs)
+        self.assertEqual(inputs.maximum_strain, 20.0)
+        self.assertEqual(inputs.duration_in_seconds, 1.0)
+        self.assertEqual(inputs.number_of_steps, 100)
```
avdg81: Perhaps we can also assert what is the default current test type (triaxial, I think)? That would also help to explain a bit the other assertions. Also, I'm curious why we're testing a subset of the test input (we are missing the initial effective cell pressure and the drainage type)? Are those less important perhaps?

kratos_element_test/view/soil_test_input_view.py
```diff
@@ -0,0 +1,313 @@
...
+        self.test_input_frame = ttk.Frame(self, padding="10")
+        self.test_input_frame.pack(fill="both", expand=True)
+
+        self._switch_test(TRIAXIAL)
+
+    def disable(self):
+        if hasattr(self, "test_type_menu") and self.test_type_menu.winfo_exists():
+            self.test_type_menu.config(state="disabled")
+
+    def _switch_test(self, test_name):
```
avdg81: This is a very long function. Perhaps in another PR we can try to extract some more functions, to make it a little bit easier to read and understand. In that way, we may also try to reuse some code, since parts of it look very similar.

kratos_element_test/view/ui_builder.py
```diff
@@ -283,168 +215,27 @@ def setup_mohr_coulomb_controls(self, params):
...
-                eps_max=eps_max,
-                n_steps=n_steps,
-                duration=duration
+                material_parameters=[float(x) for x in material_params]
             )
 
             if success:
                 self.plot_frame.draw()
+                test_type = self.controller.get_current_test_type()
+                tt = TEST_NAME_TO_TYPE.get(test_type, "triaxial")
```
avdg81: Let's move this expression to where `tt` is being used (there is only one usage a few lines down), such that we no longer introduce the variable `tt`.

kratos_element_test/view/soil_test_input_view.py
```diff
@@ -0,0 +1,313 @@
...
+        self._validate_entries_are_convertible_to_numbers(widget_dicts)
+
+    @staticmethod
+    def _validate_entries_are_convertible_to_numbers(widget_dicts):
+        for widget_dict in widget_dicts:
+            for key, widget in widget_dict.items():
+                try:
+                    float(widget.get())
+                except ValueError:
+                    raise ValueError(f"Could not convert entry to number for '{key}'.")
```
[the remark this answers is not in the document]
rfaasse: Correct. Validation deserves some more attention anyway. For now, this keeps the behavior the same as before (throwing an error when something is not a number), later I'd like to add more tailored validation using the in-built functionality

kratos_element_test/view/ui_builder.py
```diff
@@ -283,168 +215,27 @@ def setup_mohr_coulomb_controls(self, params):
...
-                test_type=tt,
                 model_name=self.model_var.get(),
                 dll_path=self.dll_path or "",
                 udsm_number=udsm_number,
-                material_parameters=[float(x) for x in material_params],
-                sigma_init=sigma_init,
-                eps_max=eps_max,
-                n_steps=n_steps,
-                duration=duration
+                material_parameters=[float(x) for x in material_params]
```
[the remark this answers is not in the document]
rfaasse: Agreed, see my previous comment

kratos_element_test/model/soil_test_input_manager.py
rfaasse: Agreed, done!

kratos_element_test/model/test/test_soil_test_input_manager.py
```diff
@@ -101,6 +101,20 @@ def test_update_init_pressure(self, test_type):
...
+
+    def test_get_current_default_inputs_for_crs(self):
+        self.input_manager.set_current_test_type("CRS")
+        inputs = self.input_manager.get_current_test_inputs()
+        self.assertIsNotNone(inputs)
+        self.assertEqual(inputs.maximum_strain, 0.0)
+        self.assertEqual(inputs.duration_in_seconds, 18000.0)
+        self.assertEqual(inputs.number_of_steps, 500)
```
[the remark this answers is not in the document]
rfaasse: The reason I tested these specifically is that they are calculated from the default strain increments. But I agree it's a bit confusing, so I extended the test

kratos_element_test/model/main_model.py
```diff
@@ -1,6 +1,48 @@
...
+        material_parameters,
+    ) -> None:
+
+        inputs = self.soil_test_input_manager.get_current_test_inputs()
+        try:
+            inputs.validate()
+        except ValueError as e:
+            self._logger("Calculation stopped due to invalid input.", "error")
+            self._logger(str(e), "error")
+            raise e
```
indigocoral:
```suggestion
        except ValueError:
            self._logger("Calculation stopped due to invalid input.", "error")
            self._logger(str(e), "error")
            raise
```

kratos_element_test/model/main_model.py
indigocoral: This is indeed very nice!
Perhaps the return annotation can be added to this one as well?
```suggestion
    def get_latest_results(self) -> Optional[Dict[str, List[float]]]:
```

rfaasse: Good point, added!

kratos_element_test/controller/element_test_controller.py
```diff
@@ -54,101 +62,50 @@ def _is_valid_test_type(self, test_type: Optional[str]) -> bool:
...
-                step_counts=getattr(self, "step_counts", None),
-                strain_incs=getattr(self, "strain_incs", None)
+            self._main_model.run_simulation(
+                model_name,
+                dll_path,
+                udsm_number,
+                mohr_coulomb_options,
+                material_parameters,
             )
+            self.latest_results = self._main_model.get_latest_results()
```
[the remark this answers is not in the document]
rfaasse: Good question, and an understandable one, because this is not the final state. For now I wanted to save the results in the model, but didn't want to change too much in the controller and view. In the next pull request I will make sure that the model has the only version of the results and remove the members in the controller and view

kratos_element_test/controller/soil_test_input_controller.py
```diff
@@ -26,6 +26,9 @@ def get_shear_inputs(self) -> TriaxialAndShearSimulationInputs:
     def get_crs_inputs(self):
         return self._soil_test_input_manager.input_data.get(CRS)
 
+    def get_current_test_inputs(self):
```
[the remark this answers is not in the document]
rfaasse: You're right, added them!

kratos_element_test/model/main_model.py
rfaasse: Good catch, before your suggestion I got the message twice:
[ERROR] Calculation stopped due to invalid input.
[ERROR] Number of steps must be > 0, but got -1.
[ERROR] Simulation failed: Number of steps must be > 0, but got -1.

Now I get:
[ERROR] Calculation stopped due to invalid input.
[ERROR] Simulation failed: Number of steps must be > 0, but got -1.

kratos_element_test/model/test/test_main_model.py
```diff
@@ -0,0 +1,23 @@
...
+
+class MainModelTest(unittest.TestCase):
+    def test_running_a_linear_elastic_triaxial_simulation(self):
+        model = MainModel(logger=lambda msg, level: None)
+
+        youngs_modulus = 9e5
+        poisson_ratio = 0.3
+        material_parameters = [youngs_modulus, poisson_ratio]
+        model.run_simulation(
+            "linear elastic model", None, 0, MohrCoulombOptions(), material_parameters
```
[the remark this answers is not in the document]
avdg81: Yes, your changes address exactly the point I was trying to make. Thank you.

avdg81 (approved):
Hi Richard,
Thanks a lot for processing the review suggestions and replying to my questions. I feel this clean and well-tested PR is good to go.

PR #44, opened 2026-01-02
Added static code analysis of pylint to the pipeline

rfaasse:
This PR adds `pylint` as static analysis tool to the pipeline. It reports best practices and warnings (see pipeline):
image

It will fail on an error (for example for non-existing imports or other things pylint thinks is an error). We could also fail on warnings if we'd like, but we first need to clean a few up before we can do that.

### DIFF
--- a/kratos_element_test/controller/element_test_controller.py
+++ b/kratos_element_test/controller/element_test_controller.py
@@ -3,9 +3,18 @@
 # Contact kratos@deltares.nl
 
 from typing import Optional, Callable, List, Tuple, Dict
+
+from kratos_element_test.controller.soil_test_input_controller import SoilTestInputController
+from kratos_element_test.model.main_model import MainModel
+from kratos_element_test.model.models import (
+    SimulationInputs,
+    MohrCoulombOptions,
+)
 from kratos_element_test.model.pipeline.run_simulation import RunSimulation
-from kratos_element_test.model.models import SimulationInputs, MohrCoulombOptions
-from kratos_element_test.view.ui_constants import VALID_TEST_TYPES, VALID_DRAINAGE_TYPES
+from kratos_element_test.view.ui_constants import (
+    VALID_TEST_TYPES,
+    VALID_DRAINAGE_TYPES,
+)
 
 
 class ElementTestController:
@@ -20,6 +29,9 @@ class ElementTestController:
 
         self._test_type: Optional[str] = None
         self._drainage: str = "drained"
+        self._main_model = MainModel()
+
+        self._soil_test_input_controller = SoilTestInputController(self._main_model.soil_test_input_manager)
 
     def set_mohr_enabled(self, enabled: bool) -> None:
         self._mc_enabled = bool(enabled)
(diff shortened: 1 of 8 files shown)


.github/workflows/workflow.yml
```diff
@@ -25,4 +25,21 @@ jobs:
       - name: Install dependencies
         run: pip install --upgrade pip .
       - name: Run tests
-        run: pytest
\ No newline at end of file
+        run: pytest
+
+  static-code-analyisis:
```
indigocoral: Small typo:
```suggestion
  static-code-analysis:
```

indigocoral (approved):
Thanks for the explanations! No further comments from my side.

kratos_element_test/model/io/udsm_parser.py
```diff
@@ -21,11 +21,11 @@ def find_symbol_in_dll(dll_path, dll_lib, symbol_name):
     try:
         pe = pefile.PE(dll_path)
         symbol_name_lower = symbol_name.lower()
-
-        for exp in pe.DIRECTORY_ENTRY_EXPORT.symbols:
-            name = exp.name
-            if name and name.decode().lower() == symbol_name_lower:
-                return getattr(dll_lib, name.decode())
+        if hasattr(pe, "DIRECTORY_ENTRY_EXPORT"):
```
avdg81: Just a question: was the original code flagged by pylint as in need of improvement? If yes, then it's fine to have this change in this PR. Otherwise, if this is an unrelated change, perhaps it's better to put it in its own PR.

.github/workflows/workflow.yml
```diff
@@ -25,4 +25,21 @@ jobs:
...
+      - uses: actions/checkout@v5
+      - name: Set up Python 3.13
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.13
+          cache: 'pip'
+      - name: Install dependencies
+        run: pip install --upgrade pip .
+      - name: Run code quality checks
+        run: pylint kratos_element_test --output-format=github --fail-under 5 --fail-on E
```
avdg81: Here we explicitly set `--fail-under 5`. When I looked up [this option in the documentation](https://pylint.pycqa.org/en/stable/user_guide/configuration/all-options.html#fail-under), it appears that the default value equals 10. Is there any particular reason why we want to deviate from the default value?

kratos_element_test/model/io/udsm_parser.py
rfaasse: Yes, this is flagged as an error by pylint, so the pipeline would fail without this change. The error is reported in the pipeline:

Image

And as you can see, the build failed: https://github.com/Deltares-research/KratosElementTestSuite/actions/runs/20659615246/job/59319160115

.github/workflows/workflow.yml
rfaasse: The reason is that we have quite some warnings right now. These will fail the pipeline when using the default. I'd like to move towards using the default, but then we first need to fix the warnings

kratos_element_test/model/test/test_result_collector.py
```diff
@@ -50,11 +50,8 @@ def test_collector_does_not_throw_when_result_file_does_not_exist(self):
             material_parameters=[1.0, 2.0, 3.0, 4.0],
             cohesion_phi_indices=(3, 4),
         )
-        try:
-            results = collector.collect_results()
-            self.assertIsInstance(results, dict)
-        except Exception as e:
-            self.fail(f"collect_results raised an exception unexpectedly: {e}")
+        results = collector.collect_results()
```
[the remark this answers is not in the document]
rfaasse: This was a pylint warning, so strictly not necessary. I reverted the change

kratos_element_test/model/io/udsm_parser.py
avdg81: Thanks for showing that without this change, an error would occur. 👍

.github/workflows/workflow.yml
avdg81: Agreed. Thanks for explaining why the default setting cannot be used yet.

PR #47, opened 2026-01-08
Model-View-Control workflow for results

rfaasse:
This PR adds a model-view-controller workflow for the results. The latest results are saved in the model (ResultManager) for each test-type and used in all places we need the results, i.e. the plots and the export.

Changelog:
- A ResultManager and a ResultController are created
- All places retrieving results are referred to the ResultManager as a single source of truth.
- Rendering the plots is moved to the plot viewer itself.

### DIFF
--- a/kratos_element_test/view/ui_constants.py
+++ b/kratos_element_test/view/ui_constants.py
@@ -7,7 +7,7 @@
 
 # Application data
 APP_TITLE = "Deltares Soil Element Test Suite"
-APP_VERSION = "Version 0.5.0 ~ Alpha Release"
+APP_VERSION = "Version 0.5.1 ~ Alpha Release"
 APP_NAME = "SoilElementSuite"
 APP_AUTHOR = "Deltares"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "kratos_element_test"
-version = "0.5.0"
+version = "0.5.1"
 description = "The Kratos Element Test Suite..."
 authors = ["Deltares Geo Kratos Team <kratos@deltares.nl>"]
 license = "Proprietary"


kratos_element_test/view/plot_viewer.py
```diff
@@ -5,44 +5,89 @@
...
-        self.canvas = FigureCanvasTkAgg(self.fig, master=self)
-        self.canvas.draw()
-        toolbar = NavigationToolbar2Tk(self.canvas, self)
+        self._grid_spec = GridSpec(
+            rows, cols, figure=self._figure, wspace=0.4, hspace=0.6
+        )
+        self._axes = [
+            self._figure.add_subplot(self._grid_spec[i]) for i in range(num_plots)
+        ]
+        self._plotter = MatplotlibPlotter(self._axes, logger=None)
```
rfaasse: For now I made this a hardcoded MatplotlibPlotter, since we also use some other matplotlib only functionality at the moment, like GridSpec, FigureCanvasTkAgg, NavigationToolbar2Tk.

kratos_element_test/model/main_model.py
```diff
@@ -33,15 +36,26 @@ def run_simulation(
         sim = RunSimulation(
             test_inputs=inputs,
             model_name=model_name,
-            dll_path=dll_path,
+            dll_path=dll_path or "",
```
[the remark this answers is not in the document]
rfaasse: This must've happened during a merge from main. Good catch thanks, reverted!

kratos_element_test/controller/result_controller.py
```diff
@@ -0,0 +1,14 @@
...
+
+class ResultController:
+    def __init__(self, result_manager: ResultManager):
+        self._result_manager = result_manager
+
+    def get_latest_results(self) -> Optional[Dict[str, List[float]]]:
+        return self._result_manager.get_results_of_active_test_type()
+
+    def get_current_test(self) -> str:
+        return self._result_manager.get_current_test()
```
indigocoral: Can it be that the `_result_manager.get_current_test` returns None, or since we always have Triaxial test as default, then it is always for sure a string?

kratos_element_test/model/result_manager.py
```diff
@@ -0,0 +1,26 @@
...
+        :param active_test_getter: callable which returns the currently active test
+        """
+        self._simulation_results: Dict[str, Dict[str, List[float]]] = {}
+        self._active_test_getter = active_test_getter
+
+    def get_results_of_active_test_type(self) -> Optional[Dict[str, List[float]]]:
+        return self._simulation_results.get(self.get_current_test())
+
+    def set_results_of_active_test_type(self, results: Dict[str, List[float]]):
+        self._simulation_results[self._active_test_getter()] = results
```
indigocoral: Question! What is the difference between these two?

kratos_element_test/view/ui_builder.py
```diff
@@ -7,50 +7,48 @@
...
 
         self.model_var = tk.StringVar(root)
         self.model_var.set(model_dict["model_name"][0])
         self.test_input_history = {}
 
         self._init_frames()
 
-        self.plot_frame = PlotViewer(self, padding="5", width=800, height=600)
+        self.plot_frame = PlotViewer(
+            self.controller._result_controller, self, padding="5", width=800, height=600
```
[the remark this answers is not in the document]
rfaasse: Well, no, not really. Right now, it's still the case that the MainUI (and by extension the UI components such as the PlotViewer) are created using a reference to the controller. 

In the end, I'd like it to be the other way around. So the controllers should be created using references to UI elements instead of the other way around. However, this logic is hard to reverse when not the entire MVC pattern is implemented, so I kept this use of an internal member as a reminder.

PR #51, opened 2026-01-20
Add functionality to use umat soil models in the suite

indigocoral:
A user would like to test a UMAT model in the KratosElementTestSuite. So far, the suite only supports reading the UDSM models input. A functionality to use UMAT soil model in the suite needs to be added.
The UMATs don't have any information about their inputs so the way to use these models in the suite is to add input fields in the suite and give the user the ability to add or remove these properties/input fields based on their UMAT DLL. These inputs will be written to the MaterialParameters.json file with the UMAT constitutive law.

### DIFF
--- a/kratos_element_test/model/pipeline/run_simulation.py
+++ b/kratos_element_test/model/pipeline/run_simulation.py
@@ -182,22 +182,24 @@ class RunSimulation:
         if self.dll_path:
             is_umat = self.udsm_number is None
 
+            editor.update_material_properties(
+                {
+                    "IS_FORTRAN_UDSM": True,
+                    "UMAT_PARAMETERS": self.material_parameters,
+                    "UDSM_NAME": self.dll_path,
+                }
+            )
             if is_umat:
                 editor.update_material_properties(
                     {
-                        "IS_FORTRAN_UDSM": True,
-                        "UMAT_PARAMETERS": self.material_parameters,
-                        "UDSM_NAME": self.dll_path,
                         "STATE_VARIABLES": [0.0]
                     }
                 )
                 editor.set_constitutive_law("SmallStrainUMAT2DPlaneStrainLaw")
+
             else:
                 editor.update_material_properties(
                     {
-                        "IS_FORTRAN_UDSM": True,
-                        "UMAT_PARAMETERS": self.material_parameters,
-                        "UDSM_NAME": self.dll_path,
                         "UDSM_NUMBER": self.udsm_number
                     }
                 )


kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -178,16 +178,32 @@ def _prepare_crs_stages(self) -> None:
...
+            else:
+                editor.update_material_properties(
+                    {
+                        "IS_FORTRAN_UDSM": True,
+                        "UMAT_PARAMETERS": self.material_parameters,
+                        "UDSM_NAME": self.dll_path,
+                        "UDSM_NUMBER": self.udsm_number
+                    }
+                )
+                editor.set_constitutive_law("SmallStrainUDSM2DPlaneStrainLaw")
```
rfaasse: There are some common changes here, maybe we can have the common changes outside of the if statements?
```suggestion
        editor.update_material_properties(
            {
                "IS_FORTRAN_UDSM": True,
                "UMAT_PARAMETERS": self.material_parameters,
                "UDSM_NAME": self.dll_path,
            }
            
            if is_umat:
                editor.update_material_properties(
                    {
                        "STATE_VARIABLES": [0.0]
                    }
                )
                editor.set_constitutive_law("SmallStrainUMAT2DPlaneStrainLaw")
            else:
                editor.update_material_properties(
                    {
                        "UDSM_NUMBER": self.udsm_number
                    }
                )
                editor.set_constitutive_law("SmallStrainUDSM2DPlaneStrainLaw")
```

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -178,16 +178,32 @@ def _prepare_crs_stages(self) -> None:
...
-            editor.update_material_properties(
-                {
-                    "IS_FORTRAN_UDSM": True,
-                    "UMAT_PARAMETERS": self.material_parameters,
-                    "UDSM_NAME": self.dll_path,
-                    "UDSM_NUMBER": self.udsm_number,
-                }
-            )
-            editor.set_constitutive_law("SmallStrainUDSM2DPlaneStrainLaw")
+            is_umat = self.udsm_number is None
```
rfaasse: For now I think this is a good enough way to check whether it's UDSM or UMAT. After refactoring the material inputs, the model will know about which type is chosen (UMAT/UDSM/Linear Elastic/Internal MC), then these if-statements get a bit simpler 👍

rfaasse (commented):
Thanks for implementing this functionality! As discussed I found a couple things during testing, but other than that, only have minor suggestions. I also see further possibility for improvement of the code, but I think it makes more sense to do that after the refactor of the flow for the material inputs

kratos_element_test/view/ui_menu.py
```diff
@@ -195,6 +196,34 @@ def load_dll():
...
+            nonlocal last_model_source
+            model_dict = {
+                "model_name": ["UMAT"],
+                "param_mode": "manual",
+                "default_param_rows": 5,
+                "manual_unit": "–",
+            }
+            dll_path = filedialog.askopenfilename(
+                title=SELECT_UMAT,
+                initialdir=_safe_udsm_initialdir(),
```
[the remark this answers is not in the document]
indigocoral: Yes yes! I already wanted to do it but I forgot! Done!

PR #53, opened 2026-01-21
Prevent removing last crs test row after running a simulation

indigocoral:
Added a helper function to prevent removing the last row in the CRS test after running a calculation.


avdg81 (commented):
I have no additional suggestions or questions. Thanks for fixing this, Marjan.

PR #64, opened 2026-01-28
Refactor material input workflow

rfaasse:
This PR moves the logic for the material inputs (either internal Mohr Coulomb, UDSM or internal linear elastic) to the model and creates a MVC workflow for the material inputs. In order to do so the following changes were made:
- A `MaterialInputView`, `MaterialInputController` and `MaterialInputManager` are created as the basis of the MVC workflow for hte material inputs
- `MaterialInputDataModels` are created as data classes which save all relevant information regarding the different materials (i.e. MC, linear elastic and UDSM)
- The data and logic in the UI as well as the controller related to material inputs has been moved to the model and unit tested. This includes loading a UDSM.

### DIFF
--- a/kratos_element_test/model/material_input_manager.py
+++ b/kratos_element_test/model/material_input_manager.py
@@ -58,7 +58,7 @@ class MaterialInputManager:
         self._material_inputs["udsm"].clear()
         self._current_udsm_index = 0
         model_dict = udsm_parser(str(dll_path.resolve()))
-        for index, (parameter_names, parameter_units, model_name) in enumerate(
+        for udsm_number, (parameter_names, parameter_units, model_name) in enumerate(
             zip(
                 model_dict["param_names"],
                 model_dict["param_units"],
@@ -71,7 +71,7 @@ class MaterialInputManager:
                 user_defined_parameters[name] = Parameter(0.0, unit)
             inputs = UDSMMaterialInputs()
             inputs.material_parameters["UDSM_NAME"] = str(dll_path.resolve())
-            inputs.material_parameters["UDSM_NUMBER"] = index
+            inputs.material_parameters["UDSM_NUMBER"] = udsm_number
             inputs.user_defined_parameters = user_defined_parameters
             inputs.model_name = model_name
             self._material_inputs["udsm"].append(inputs)


kratos_element_test/view/ui_builder.py
```diff
@@ -166,113 +114,15 @@ def _create_input_fields(self):
...
-        elif self.is_mohr_coulomb:
-            self.controller.set_mohr_enabled(True)
-
-            c_idx, phi_idx = self._parse_mc_indices()
-            self.controller.set_mohr_mapping(c_idx, phi_idx)
-
-            self.mohr_frame.pack_forget()
-
-        else:
-            self.mohr_checkbox_widget.configure(state="normal")
```
rfaasse: This functionality is moved to the MaterialInputView

kratos_element_test/controller/material_input_controller.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class MaterialInputController:
+    def __init__(self, material_input_manager: MaterialInputManager):
+        self._material_input_manager = material_input_manager
+
+    def get_current_material_inputs(self) -> LinearElasticMaterialInputs:
+        return self._material_input_manager.get_current_material_inputs()
+
+    def bind_test_input_fields_to_update_functions(self, string_vars) -> None:
```
indigocoral: If I understand correctly, this is material input, right? so maybe renaming it to something like `bind_material_fields...` would be better?

indigocoral (commented):
Thanks a lot, Richard, for doing this refactoring of moving the material input logic.
So far, I have these comments. But I think I would still continue looking at it more tomorrow or so. I also have a remark for how the selection model is in the UI now, which I will show it to you in person.
But overall nice and understandable PR and I have tested it and haven't seen anything unusual! So thanks! :)

kratos_element_test/controller/material_input_controller.py
rfaasse: Indeed, done!

kratos_element_test/model/material_input_manager.py
```diff
@@ -0,0 +1,113 @@
...
+    Parameter,
+)
+
+
+class MaterialInputManager:
+    def __init__(self):
+        self._current_material_type = ""
+        self._material_inputs = {
+            "linear_elastic": LinearElasticMaterialInputs(),
+            "mohr_coulomb": MohrCoulombMaterialInputs(),
```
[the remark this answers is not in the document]
rfaasse: If I understand correctly, that depends if we'd like to keep a history if inputs, such that we can add it to a list. Right now I only have one input per 'key', so I kept it a singular object

kratos_element_test/model/material_input_data_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    ),
+) -> tuple[None, None] | tuple[float, float]:
+    try:
+        return material_inputs.get_cohesion_and_phi()
+    except AttributeError:
+        return None, None
+
+
+def convert_user_inputs_to_kratos_inputs(
+    user_defined_parameters: Dict[str, float],
```
avdg81: If I'm not mistaken, the values of the dict `user_defined_parameters` are of type `Parameter` rather than `float`:
```suggestion
    user_defined_parameters: Dict[str, Parameter],
```

indigocoral (approved):
Thanks a lot for addressing the review comments and answering my questions. And also for fixing the space in the UI. Nothing more from my side. 👍🏼

kratos_element_test/model/material_input_data_models.py
```diff
@@ -0,0 +1,106 @@
...
+        cohesion = self.user_defined_parameters["GEO_COHESION"].value
+        phi = self.user_defined_parameters["GEO_FRICTION_ANGLE"].value
+        return cohesion, phi
+
+
+@dataclass
+class UDSMMaterialInputs:
+    kratos_law_name: str = "SmallStrainUDSM2DPlaneStrainLaw"
+    model_name: str = ""
+    user_defined_parameters = Dict
```
avdg81: I'm a little confused here: shouldn't `Dict` just be used a type hint, and then `{}` as initialization? I haven't seen it written this way in the other data classes. (I know very little about the `typing` module, so if this is common practice, please teach me).
```suggestion
    user_defined_parameters: Dict = {}
```

kratos_element_test/model/test/test_material_input_manager.py
```diff
@@ -0,0 +1,166 @@
...
+        material_input_manager = MaterialInputManager()
+        self.assertRaises(
+            ValueError,
+            lambda: material_input_manager.set_current_material_type("nonexistent"),
+        )
+
+    def test_loading_udsm_initializes_material_inputs(self):
+        material_input_manager = MaterialInputManager()
+        material_input_manager.initialize_udsm(
+            Path(soil_models_dir()) / "sclay1creep.dll"
```
avdg81: The following suggestion is unrelated to the work you've done for this PR, so perhaps it can be addressed another time. Wouldn't it be more logical if `soil_models_dir` returns a `Path` instance? Now it returns a string (whereas internally, it already uses `Path` objects, but converts it to a string on `return`), which forces you to explicitly convert back to a `Path` object here. I would only convert paths to strings when that's absolutely necessary. Otherwise, `Path` objects communicate better what they represent than strings do.

kratos_element_test/model/test/test_material_input_manager.py
```diff
@@ -0,0 +1,166 @@
...
+            "n": Parameter(value=0.0, unit="-"),
+            "fₜ𝒸": Parameter(value=0.0, unit="Degrees"),
+            "r": Parameter(value=0.0, unit="-"),
+            "w": Parameter(value=0.0, unit="-"),
+            "w𝒹": Parameter(value=0.0, unit="-"),
+            "t": Parameter(value=0.0, unit="days"),
+            "a₀": Parameter(value=0.0, unit="-"),
+            "OCR": Parameter(value=0.0, unit="-"),
+            "e₀": Parameter(value=0.0, unit="-"),
+        }
```
avdg81: FYI: I don't really know how to check these expectations, so I just assume these are correct.

kratos_element_test/model/test/test_material_input_manager.py
```diff
@@ -0,0 +1,166 @@
...
+        self.assertEqual(
+            udsm_material_inputs.user_defined_parameters,
+            expected_changeable_material_parameters,
+        )
+        self.assertEqual(
+            udsm_material_inputs.get_kratos_inputs()["UMAT_PARAMETERS"],
+            [0.0] * len(expected_changeable_material_parameters),
+        )
+
+        material_input_manager.set_current_udsm_number(1)
```
avdg81: I'm a bit confused here: wasn't it already set to 1? (See also the assertion at line 92.)

Edit: I only later realized that we're setting the **index** here (which is off by one compared to the UDSM number that the user provides). I feel that a clear distinction between number and index may help here.

kratos_element_test/model/test/test_material_input_manager.py
```diff
@@ -0,0 +1,166 @@
...
+            "w𝒹": Parameter(value=0.0, unit="-"),
+            "t": Parameter(value=0.0, unit="days"),
+            "a₀": Parameter(value=0.0, unit="-"),
+            "OCR": Parameter(value=0.0, unit="-"),
+            "e₀": Parameter(value=0.0, unit="-"),
+            "Efib": Parameter(value=0.0, unit="kN/m"),
+            "sₜₑₙ": Parameter(value=0.0, unit="kN/m"),
+            "s𝒸ₒₘ": Parameter(value=0.0, unit="kN/m"),
+            "a": Parameter(value=0.0, unit="[Degrees]"),
+        }
```
avdg81: Also here, I don't know how I can check these expected values.

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -32,30 +38,25 @@
...
         *,
         test_inputs: TriaxialAndShearSimulationInputs | CRSSimulationInputs,
-        model_name: Optional[str],
-        dll_path: Optional[str],
-        udsm_number: Optional[int],
-        material_parameters: List[float],
-        cohesion_phi_indices: Optional[Tuple[int, int]] = None,
+        material_inputs: (
+            LinearElasticMaterialInputs | MohrCoulombMaterialInputs | UDSMMaterialInputs
+        ),
```
avdg81: This function has become much cleaner now 👍

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -113,9 +114,8 @@ def post_process_results(self) -> Dict[str, List[float]]:
             self.log("Collecting results...", "info")
 
             output_file_strings = [str(p) for p in self._output_file_paths()]
-            collector = ResultCollector(
-                output_file_strings, self.material_parameters, self.cohesion_phi_indices
-            )
+            cohesion, phi = get_cohesion_and_phi(self.material_inputs)
+            collector = ResultCollector(output_file_strings, cohesion, phi)
```
avdg81: Rather than passing the cohesion and phi to the `ResultCollector`, we may want to insert these two properties in the dict returned by `collect_results`. In this way, the `ResultCollector` doesn't need to know anything about these two input parameters. But again, not for this PR. Just so you can make a note of it, and return to it later. 👍

kratos_element_test/view/ui_builder.py
```diff
@@ -93,11 +73,10 @@ def _init_frames(self):
         )
         self.scroll_canvas.bind_all("<MouseWheel>", self._on_mousewheel)
 
-        self.dropdown_frame = ttk.Frame(self.left_frame)
-        self.dropdown_frame.pack(fill="x")
-
-        self.param_frame = SoilParameterEntries(self.left_frame, padding="10")
-        self.param_frame.pack(fill="both", expand=True, pady=10)
+        self.material_input_view = MaterialInputView(
+            self.controller._material_input_controller, self.left_frame, padding="10"
```
avdg81: Just a remark: here we're using a private attribute directly. Perhaps we want to reconsider that later? For now, no action is required, but in general, client code shouldn't be using private implementation details directly.

kratos_element_test/model/material_input_manager.py
```diff
@@ -0,0 +1,113 @@
...
+    def get_current_material_inputs(
+        self,
+    ) -> LinearElasticMaterialInputs | MohrCoulombMaterialInputs | UDSMMaterialInputs:
+        if (
+            self.get_current_material_type() == "udsm"
+            and len(self._material_inputs["udsm"]) == 0
+        ):
+            raise RuntimeError(
+                "UDSM material inputs have not been initialized. Please initialize UDSM first."
+            )
```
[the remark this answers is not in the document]
rfaasse: No, in the GUI the initialize_udsm function is always called before asking for the current UDSM material inputs for the first time, so it's indeed a programmer's error.

kratos_element_test/model/material_input_manager.py
```diff
@@ -0,0 +1,113 @@
...
+
+        return (
+            self._material_inputs[self.get_current_material_type()][
+                self._current_udsm_number
+            ]
+            if self.get_current_material_type() == "udsm"
+            else self._material_inputs[self.get_current_material_type()]
+        )
```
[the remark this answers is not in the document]
rfaasse: Agreed, this is clearer!

kratos_element_test/model/material_input_manager.py
```diff
@@ -0,0 +1,113 @@
...
+            self._material_inputs["udsm"].append(inputs)
+
+    def set_current_udsm_number(self, udsm_number):
+        self._current_udsm_number = udsm_number
+
+    def get_udsm_model_names(self):
+        if len(self._material_inputs["udsm"]) == 0:
+            raise RuntimeError(
+                "UDSM material inputs have not been initialized. Please initialize UDSM first."
+            )
```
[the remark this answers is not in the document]
rfaasse: This should never happen through the GUI, it's the programmer's responsibility. Changed to an assert for clarity

kratos_element_test/model/material_input_manager.py
```diff
@@ -0,0 +1,113 @@
...
+        return [udsm_inputs.model_name for udsm_inputs in self._material_inputs["udsm"]]
+
+    def set_current_udsm_model(self, model_name):
+        self._current_udsm_number = self.get_udsm_model_names().index(model_name)
```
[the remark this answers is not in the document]
rfaasse: Good idea, done!

kratos_element_test/model/material_input_manager.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def set_phi_index(self, phi_index):
+        if self.get_current_material_type() == "udsm":
+            material_inputs = self.get_current_material_inputs()
+            material_inputs.mohr_coulomb_options.phi_index = phi_index
+
+    def get_mohr_enabled(self):
+        if self.get_current_material_type() == "udsm":
+            material_inputs = self.get_current_material_inputs()
+            return material_inputs.mohr_coulomb_options.enabled
```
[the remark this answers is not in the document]
rfaasse: Good catch, used and assert instead of the conditional, since it should never happen this is called when the type is not udsm

kratos_element_test/model/test/test_material_input_manager.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+
+        linear_elastic_material_inputs = (
+            material_input_manager.get_current_material_inputs()
+        )
+
+        self.assertEqual(
+            linear_elastic_material_inputs.get_kratos_inputs()["YOUNG_MODULUS"],
+            9e5,
+        )
```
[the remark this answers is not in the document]
rfaasse: Good point, missed this one

kratos_element_test/model/test/test_material_input_manager.py
rfaasse: The only way is to take the current master (e.g. via the pip install route described in the readme https://github.com/Deltares-research/KratosElementTestSuite), import the shipped sclay1creep.dll and check if the labels and units are the same.

kratos_element_test/model/test/test_material_input_manager.py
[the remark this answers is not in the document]
rfaasse: It can't happen, so changed it to an assert

kratos_element_test/model/test/test_material_input_manager.py
rfaasse: See previous answer. I'm not sure if it's really necessary, but that's the way to check it if you want to

PR #66, opened 2026-02-02
Add undrained material behaviour for Direct Simple Shear test

indigocoral:
This PR enables Direct Simple Shear (DSS) undrained tests to run using the base drained templates by programmatically setting the "IGNORE_UNDRAINED" flag to false and unfixing the water pressures in the DSS test for the undrained test. This removes the requirement for a duplicate "undrained" template folder for DSS tests; thus, I removed that part of the code I wrote for it. Also, a plot function was added to plot the excess pore pressure vs. shear strain for the undrained DSS test.

### DIFF
--- a/kratos_element_test/controller/soil_test_input_controller.py
+++ b/kratos_element_test/controller/soil_test_input_controller.py
@@ -4,6 +4,7 @@ from kratos_element_test.model.models import (
     CRSSimulationInputs,
 )
 from kratos_element_test.model.soil_test_input_manager import SoilTestInputManager
+from kratos_element_test.view.ui_logger import log_message
 from kratos_element_test.view.ui_constants import (
     TRIAXIAL,
     DIRECT_SHEAR,
@@ -109,6 +110,8 @@ class SoilTestInputController:
             ):
                 num_plots = 5 if val == "undrained" else 4
                 on_drainage_changed(num_plots=num_plots)
+
+                log_message(f"{val} test type selected.", "info")
 
         combo_box.bind("<<ComboboxSelected>>", _sync_drainage_from_combobox)
 


kratos_element_test/model/pipeline/result_collector.py
```diff
@@ -223,6 +275,19 @@ def _apply_cumulative_strain_offset(
...
+        all_water_pressures = []
+        for path in result_paths:
+            _, _, _, _, _, wp, _ = self._read_results(path)
+            water_pressure_values = self._extract_water_pressure(wp)
+            all_water_pressures.extend(water_pressure_values)
+
+        if not all_water_pressures:
+            return []
+
+        initial_pressure = all_water_pressures[0]
```
rfaasse: As mentioned before, I'd keep the reading of the file in the 'general' part (such that we do it only once)

kratos_element_test/view/soil_test_input_view.py
```diff
@@ -213,22 +215,48 @@ def _switch_test(self, test_name):
...
+        )
+
+    def _on_drainage_change(self, event=None):
+        current_test = self._soil_test_input_controller.get_current_test_type()
+        val = self.test_type_var.get().strip().lower()
+        is_undrained = val == "undrained"
+
+        if current_test == DIRECT_SHEAR:
+            num_plots = 5 if is_undrained else 4
+            self.update_plots_callback(num_plots=num_plots)
```
rfaasse: If I understand the flow correctly I don't think we need this _on_drainage_change function, since we already do a bind in the soilt_test_input_controller (the _sync_drainage_from_combobox function). Would it be possible to add the extra functionality (so the updating of the plots etc) to that function and remove this one?

kratos_element_test/model/pipeline/run_simulation.py
```diff
@@ -221,6 +225,11 @@ def _set_material_constitutive_law(self) -> None:
                 )
                 editor.set_constitutive_law("GeoMohrCoulombWithTensionCutOff2D")
 
+    def _set_material_undrained(self) -> None:
+        if self.drainage and str(self.drainage).strip().lower() == "undrained":
+            editor = MaterialEditor(str(self.material_json_path))
+            editor.set_undrained_flag(False)
```
rfaasse: I think next to setting the undrained flag, we need to 'unfix' the water pressure in the ApplyScalarConstraintTableProcess in the `ProjectParameters.json`. Right now, the water pressure can never become anything else than 0.0 if I read the template files correctly

rfaasse (commented):
Thanks for implementing this. I've tested it and it runs (although I think we are missing a change in the project parameters to make sure the water pressures are free, which if I understand correctly is needed for undrained tests, but correct me if I'm wrong).

Next to that, I have some more suggestions on a code level (some simplifications, adding a test), but nothing major!

indigocoral:
As discussed and pointed out by @WPK4FEM , dilatancy angle has an effect on the DSS test in building up pore pressure, which we don't get right now, unfortunately. We need to wait for the undrained functionality to be implemented in the GeoMechanicsApplication and then bring this back to master.

PR #73, opened 2026-03-01
Visualize the experimental data on the plots from Python

indigocoral:


kratos_element_test/controller/element_test_controller.py
```diff
@@ -56,6 +57,41 @@ def export_latest_results(self):
             raise ValueError("No results available for export")
         export_excel_by_test_type(results, test_type)
 
+    def _import_lab_results(self, py_file: Path) -> None:
+        spec = importlib.util.spec_from_file_location(
```
rfaasse: What does `spec` stand for? I see it's an object of ModuleSpec type, is it a module specifier/specification?

kratos_element_test/controller/element_test_controller.py
```diff
@@ -56,6 +57,41 @@ def export_latest_results(self):
...
+                continue
+            if not isinstance(results, dict) or not results:
+                continue
+
+            storage_key = type_to_name.get(test_type, test_type)
+            self._result_controller.set_experimental_results_for_test_type(
+                storage_key, results
+            )
+
+        self._logger(f"Imported lab results from {py_file}", "info")
```
rfaasse: Since there is no GUI interaction here and the path to the python file is already known, I think we could move this functionality/logic to the result_manager (or at least a function in the model lib), such that we can test it automatically :)

In this function we could then simply call a method in the self._main_model which forwards the call to the result manager (which then imports the experimental results directly to its own member). Maybe we can have a look together at how to get there 👍

kratos_element_test/controller/element_test_controller.py
```diff
@@ -56,6 +57,41 @@ def export_latest_results(self):
...
+        if experimental_by_test is None:
+            experimental_by_test = getattr(module, "experimental", None)
+
+        if not isinstance(experimental_by_test, dict) or not experimental_by_test:
+            raise ValueError(
+                "No 'experimental_by_test' (or 'experimental') dict found in lab results file"
+            )
+
+        type_to_name = {v: k for k, v in TEST_NAME_TO_TYPE.items()}
+
```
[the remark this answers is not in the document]
indigocoral: Good point. I added it. Let me know if this solves it!

kratos_element_test/view/ui_builder.py
```diff
@@ -136,6 +136,13 @@ def _run_simulation(self):
             self.root.after(0, self._enable_gui)
             self.is_running = False
 
+    def refresh_plots(self) -> None:
```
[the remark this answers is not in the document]
indigocoral: Done

kratos_element_test/plotters/lab_result_overlay_registry.py
```diff
@@ -0,0 +1,242 @@
...
+    TITLE_VERTICAL_STRESS_VS_HORIZONTAL_STRESS,
+)
+
+Transform = Callable[[List[float]], List[float]]
+ComputeXY = Callable[
+    [Dict[str, List[float]]], Optional[Tuple[List[float], List[float]]]
+]
+
+
+@dataclass(frozen=True)
```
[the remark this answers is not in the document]
indigocoral: Yes, exactly. `frozen=True` makes the dataclass effectively immutable after construction: you can set the fields when creating the instance, but assigning to them later raises an error. I used it here because OverlaySpec is configuration data and should not be mutated at runtime.

kratos_element_test/controller/element_test_controller.py
indigocoral: Agreed. This logic is not controller/UI-specific, so I moved it into the model layer. Now ElementTestController.import_lab_results() has become a thin forwarder to MainModel, which forwards to ResultManager.

kratos_element_test/model/test/test_result_collector.py
```diff
@@ -102,6 +107,30 @@ def test_collector_appends_results_of_multiple_stages(self):
...
+        sigma_3 = [50.0, 75.0, 100.0]
+
+        sigma, tau = compute_mohr_circle_xy(sigma_1, sigma_3, n_points=3)
+
+        expected_sigma = [200.0, 150.0, 100.0]
+        expected_tau = [0.0, -50.0, 0.0]
+
+        np.testing.assert_array_almost_equal(sigma, expected_sigma)
+        np.testing.assert_array_almost_equal(tau, expected_tau)
+
```
rfaasse: Since these are not really testing the result collector, but are more focused on the core utils, I'd propose to create a new file, `test_core_utils.py` and move the tests there

PR #76, opened 2026-03-16
Increased the number of max_iterations to 50 for convergence issues

indigocoral:


rfaasse (approved):
Good idea, 50 seems like a more reasonable number! Let's merge this and test it for some time, see if we encounter any more issues with unexpected non-convergence

PR #77, opened 2026-03-19
[ElementTest] Add possibility to import CSV to visualize the lab results

### LINKED ISSUE #74: [ElementTest] Add possibility to import CSV to visualize the lab results
Add support for importing laboratory test results from CSV files into the ElementTest GUI. The CSV data should be mapped to the expected result variables and integrated with the existing functionality so the imported results can be visualized in the current plotting workflow.
Defining the CSV format is also within this issue.
If there is no need for the developed Python workflow for visualization of the lab results, remove it.

An example of the CSV would be:

[SCLAY1Sresults.xlsx](https://github.com/user-attachments/files/26025049/SCLAY1Sresults.xlsx)

The results labels are defined in example_lab_results.py in the plotter folder. But we probably need to add more labels to the dictionary so that if the users define it differently, the plotting can still work.
We still need to ask the users how they would define such lab results in a CSV and with what label.

markelov208:
Added import of CSV  file with data. The features: 
- columns in the file can be separated also with `;`
- the file may hold Greek symbols which will be replaced with Latin letters for processing. 
- the import feature checks the variable names in the file and test and pre-fills the map when there are matches
- popup window provides a map to be filled by the user: test variable names vs column names from the file
- if the user chose the same column more than once, a warning message is printed in Log Output window and the second assignment in the map is applied
- data is used for a chosen test (default is Triaxial), other tests use their own data


kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
...
+                selected_file,
+                column_mapping=column_mapping,
+                target_test_type=current_test_internal_name,
+            )
+
+            if self.main_frame:
+                self.main_frame.redraw_plots()
+
+        except Exception as e:
+            messagebox.showerror("Import Error", f"Failed to import CSV data.\n\n{e}")
```
indigocoral, line 427: I don't think all of this belongs to here in `ui_menu.py`... I think the main logic of the CSV import function should be moved to the `result_manager.py`. So parts of `_import_csv_data` like the
`active test type` can be moved to the controller.

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
...
+    def _import_csv_data(self):
+        try:
+            current_test_display_name = self._controller.get_current_test_type()
+            if not current_test_display_name or current_test_display_name.strip() == "":
+                messagebox.showerror(
+                    "Import Error",
+                    "No active test selected. Please select a test (Triaxial, Direct Simple Shear, or CRS) "
+                    "before importing CSV data.",
+                )
+                return
```
indigocoral, line 388: This new guard has unintentionally changed behavior. Previously, after selecting a material model, Triaxial was the default test. Now the CSV import depends on get_current_test_type() being non-empty, so if the default test is no longer set automatically elsewhere, the user is forced to manually choose a test first. And if they don't want to use the import functionality, they still are forced to choose a test type first and then continue... I think I liked that the triaxial test was chosen as a default. Is it possible to keep that?

main.py
```diff
@@ -0,0 +1,24 @@
...
+        ui.create_menu()
+    except ImportError as e:
+        print(
+            f"[ERROR] Could not run GUI: {e}. Make sure the GeoMechanicsApplication is built and installed correctly."
+        )
+        sys.exit(1)
+
+
+if __name__ == "__main__":
+    main()
```
indigocoral: I don't fully understand why this file was created because we have this file in the view folder. Could you perhaps elaborate on why this was necessary? I think maybe you wanted to fix the error that I was seeing?
Maybe we can discuss if this piece of code that is for running the application actually belongs here or in the view folder.

kratos_element_test/view/kratos_element_test_gui.py
```diff
@@ -4,6 +4,15 @@
...
 import sys
+from pathlib import Path
+
+# Allow running this file directly without requiring python -m.
+if __package__ in (None, ""):
+    package_parent = Path(__file__).resolve().parents[2]
+    package_parent_str = str(package_parent)
+    if package_parent_str not in sys.path:
+        sys.path.insert(0, package_parent_str)
+
```
indigocoral, line 15: Maybe you intended this change for the running issue I was seeing as well? This file is the one you have copied to the outside of this folder (without this change). Let's keep one only.

kratos_element_test/view/soil_test_input_view.py
```diff
@@ -73,7 +73,12 @@ def __init__(
         self.test_input_frame = ttk.Frame(self, padding="10")
         self.test_input_frame.pack(fill="both", expand=True)
 
-        self._switch_test(TRIAXIAL)
+        # Add placeholder label
+        ttk.Label(
+            self.test_input_frame,
+            text="Please select a test (Triaxial, Direct Simple Shear, or CRS) to get started.",
+            font=(HELP_MENU_FONT, 11),
+        ).pack(anchor="center", pady=(50, 0))
```
indigocoral: Please see my previous comment about keeping the default test.

kratos_element_test/view/ui_builder.py
```diff
@@ -143,6 +143,9 @@ def redraw_plots(self) -> None:
             log_message("Failed to refresh plots:", "error")
             log_message(traceback.format_exc(), "error")
 
+    def select_test(self, test_name: str) -> None:
+        self.soil_test_input_view._switch_test(test_name)
```
indigocoral: Is this being used anywhere?

kratos_element_test/view/ui_menu.py
```diff
@@ -3,14 +3,46 @@
...
+            return str(app_dir)
+
+        if sys.platform == "darwin":
+            return str(Path.home() / "Library" / "Application Support" / appname)
+
+        xdg_data_home = os.environ.get("XDG_DATA_HOME")
+        base_dir = (
+            Path(xdg_data_home) if xdg_data_home else Path.home() / ".local" / "share"
+        )
+        return str(base_dir / appname)
```
indigocoral, line 54: I'm reading this and I think this is the fix for my package error. Nice thanks!
My only request is to keep all the imports at the top of the file and then go on to functions and classes and methods. This way we avoid duplicate imports, and when we scroll to the top of the file, we can always check more easily what things are being imported and what not.

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "q",
+    ],
+    "crs": [
+        "yy_strain",
+        "sigma_yy",
+        "sigma_xx",
+        "p'",
+        "q",
+        "time_steps",
+    ],
```
indigocoral, line 203: "sigma_3" and "sigma_1" are missing from this list. Thus we don't get the Mohr circle plot.
```suggestion
    "crs": [
        "yy_strain",
        "sigma_yy",
        "sigma_xx",
        "sigma_3",
        "sigma_1",
        "p'",
        "q",
        "time_steps",
    ],
```

kratos_element_test/controller/element_test_controller.py
```diff
@@ -51,16 +48,34 @@ def get_current_test_type(self) -> str:
         return self._main_model.get_current_test_type()
 
     def export_latest_results(self):
+        from kratos_element_test.view.result_exporter import export_excel_by_test_type
```
rfaasse: To me it's not directly clear why this import is moved to this function. I think in general we do the import at the top of the file, unless there's another reason (e.g. I've moved it once due to a circular import issue).

main.py
rfaasse, line 1: I it's a good idea to rename this to main.py to clarify what its use is!

kratos_element_test/controller/element_test_controller.py
```diff
@@ -51,16 +48,34 @@ def get_current_test_type(self) -> str:
...
+        column_mapping: Optional[Dict[str, str]] = None,
+        target_test_type: Optional[str] = None,
+    ) -> str:
+        test_type_used = self._main_model.import_csv_data(
+            csv_file,
+            column_mapping,
+            target_test_type=target_test_type,
+        )
+        self._logger(f"Imported CSV data from {csv_file}", "info")
+        return test_type_used
```
rfaasse: Looking at the usage of this function, the return value is not used, so I think we can remove it.

kratos_element_test/model/result_manager.py
```diff
@@ -71,4 +77,46 @@ def import_python_lab_results(self, py_file: Path) -> None:
...
+        target_test_type: Optional[str] = None,
+    ) -> str:
+        current_test = self.get_current_test()
+        if not current_test or current_test.strip() == "":
+            raise ValueError(
+                "No active test selected. Please select a test (Triaxial, Direct Simple Shear, or CRS) "
+                "before importing CSV data."
+            )
+
+        selected_test = target_test_type or current_test
```
rfaasse, line 130: Since the current test type is always populated at this point in the function (otherwise it'd have thrown already), it is unclear to me why we need the target_test_type parameter.

kratos_element_test/model/main_model.py
```diff
@@ -63,3 +70,15 @@ def initialize_udsm(self, dll_path: Path):
...
     def import_lab_results(self, py_file: Path) -> None:
         self._result_manager.import_python_lab_results(py_file)
+
+    def import_csv_data(
+        self,
+        csv_file: Path,
+        column_mapping: Dict[str, str] | None = None,
+        target_test_type: str | None = None,
+    ) -> str:
+        return self._result_manager.import_csv_lab_results(
```
rfaasse: By extension of [this comment](https://github.com/Deltares-research/KratosElementTestSuite/pull/77/changes#r2980111071), I don't think we need the return value here

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
                 return
 
             self._controller.import_lab_results(Path(py_path))
-            self.main_frame.redraw_plots()
+            if self.main_frame:
+                self.main_frame.redraw_plots()
```
rfaasse, line 372: Is it realistic here that the main frame doesn't exist? If not, I would propose to omit the check

kratos_element_test/view/soil_test_input_view.py
rfaasse: To me it's not completely clear why we can't have the triaxial test as default anymore due to the csv import. I would prefer to keep the default if it's possible, but maybe I've missed something here

kratos_element_test/view/ui_builder.py
```diff
@@ -143,6 +143,9 @@ def redraw_plots(self) -> None:
             log_message("Failed to refresh plots:", "error")
             log_message(traceback.format_exc(), "error")
 
+    def select_test(self, test_name: str) -> None:
```
rfaasse: It might be a mistake in my CLion, but it seems this function is unused (and can be removed)

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "yy_strain",
+        "vertical_strain",
+        "axial_strain",
+        "axial strain",
+        "axial deformation",
+        "strain_yy",
+        "epsilon_yy",
+        "eps_yy",
+        "eyy",
+        "epsilon_yy_percent",
```
indigocoral: This is by percent which is not what we are supporting so I think it's best to remove this:

```suggestion
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    for canonical_key, aliases in raw_aliases.items():
+        for alias in aliases:
+            alias_map[_normalize_token(alias)] = canonical_key
+    return alias_map
+
+
+_RAW_COLUMN_ALIASES = {
+    "yy_strain": (
+        "yy_strain",
+        "vertical_strain",
```
indigocoral, line 49: Maybe adding one without the underline is also good
```suggestion
        "vertical_strain",
        "vertical strain",
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "axial_strain",
+        "axial strain",
+        "axial deformation",
```
indigocoral: Deformation is not the same thing as strain. I would remove this one.

```suggestion
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+) -> Dict[str, str]:
+    alias_map: Dict[str, str] = {}
+    for canonical_key, aliases in raw_aliases.items():
+        for alias in aliases:
+            alias_map[_normalize_token(alias)] = canonical_key
+    return alias_map
+
+
+_RAW_COLUMN_ALIASES = {
+    "yy_strain": (
```
indigocoral: I think it's best to show the user the label we use in the plotting also because it would be much clearer for the users:
```suggestion
    "Vertical Strain": (
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "axial deformation",
+        "strain_yy",
+        "epsilon_yy",
+        "eps_yy",
+        "eyy",
+        "epsilon_yy_percent",
+        "strain",
+        "vertical strain eyy",
+    ),
+    "vol_strain": (
```
indigocoral:
```suggestion
    "Volumetric Strain": (
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "volumetric strain ev",
+    ),
+    "sigma_1": (
+        "sigma_1",
+        "sigma1",
+        "s1",
+        "principal_stress_1",
+        "major_principal_stress",
+        "axial_stress",
+        "sigma1effective",
```
indigocoral: I would replace this with:
```suggestion
        "first_principal_stress",
        "first principal stress",
        "major principal stress",
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    "sigma_3": (
+        "sigma_3",
+        "sigma3",
+        "s3",
+        "principal_stress_3",
+        "minor_principal_stress",
+        "confining_stress",
+        "confining_pressure",
+        "cell_pressure",
+        "sigma3effective",
```
indigocoral: I would replace this with:
```suggestion
        "third_principal_stress",
        "third principal stress",
        "minor principal stress",
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    ),
+    "sigma1_sigma3_diff": (
+        "sigma1_sigma3_diff",
+        "delta_sigma",
+        "deltasigma",
+        "sigma1_minus_sigma3",
+        "q_diff",
+        "deviator_sigma",
+    ),
```
indigocoral: I see you have already nicely put the `_compute_missing_sigma_diff` function in lab_results_csv_parser.py. Thank you! If we are using it all the time to compute the difference between sigma1 and sigma3 (as rightly so because the user won't be giving us the diff), then I think we can remove this completely.

indigocoral (commented):
Thank you so much Gennady for all the hard work you have done on this. You have implemented it nicely. I've put my review comments on it. Although they are not finished yet, I'm forced to restart my laptop now, so I'll submit these so I don't lose them, and then I will continue the review. Thank you very much!

kratos_element_test/model/test/test_soil_test_input_manager.py
```diff
@@ -103,7 +103,15 @@ def test_update_init_pressure(self, test_type):
...
-        self.assertEqual(self.input_manager.get_current_test_type(), TRIAXIAL)
+        # No test is selected by default, so get_current_test_type() returns empty string
+        self.assertEqual(self.input_manager.get_current_test_type(), "")
+        # Trying to get inputs when no test is selected should raise an error
+        with self.assertRaises(ValueError) as context:
+            self.input_manager.get_current_test_inputs()
+        self.assertIn("No test type has been selected", str(context.exception))
+
+        # After setting a test type, get_current_test_inputs() should work
+        self.input_manager.set_current_test_type(TRIAXIAL)
```
rfaasse: As mentioned before, I think the default of triaxial still makes sense to keep

kratos_element_test/model/main_model.py
```diff
@@ -21,6 +21,13 @@ def get_current_test_type(self) -> str:
 
     def run_simulation(self) -> None:
 
+        from kratos_element_test.model.pipeline.run_simulation import RunSimulation
```
rfaasse: Similar question as before: what is the reason we moved this to a local import instead of at the top of the file?

kratos_element_test/model/material_input_manager.py
rfaasse, line 1: The change in this file seems unrelated. In my opinion we should revert it here and move it to a different PR if needed. Personally, I think this change is not really necessary, since we specify pefile as a requirement in the toml file

kratos_element_test/plotters/lab_result_overlay_registry.py
```diff
@@ -46,7 +46,7 @@ class OverlaySpec:
     axis_index: int
     x_key: Optional[str] = None
     y_key: Optional[str] = None
-    label: str = "Lab results"
+    label: str = "Experimental"
```
rfaasse: I'm not sure about this change in label. I think it's unrelated and I think the old label was slightly more. I don't know what your and @indigocoral's opinions are?

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    "p'": (
+        "p'",
+        "pprime",
+        "p_prime",
+        "p",
```
indigocoral: I would remove this, as this can also mean water pressure for many...

kratos_element_test/model/result_manager.py
```diff
@@ -71,4 +77,46 @@ def import_python_lab_results(self, py_file: Path) -> None:
...
+        )
+
+        selected_test_results = experimental_by_test.get(effective_target_test_type, {})
+        if not selected_test_results:
+            selected_display_name = TYPE_TO_TEST_NAME.get(
+                effective_target_test_type, effective_target_test_type
+            )
+            raise ValueError(
+                f"CSV does not contain data for the selected test '{selected_display_name}'."
+            )
```
rfaasse: Since effective_target_test_type was the 'TEST_NAME_TO_TYPE.get(selected_test, selected_test)', when converting this back I think it'd return to `selected_test` again right?
```suggestion
            raise ValueError(
                f"CSV does not contain data for the selected test '{selected_test}'."
            )
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    ),
+    "sigma1_sigma3_diff": (
+        "sigma1_sigma3_diff",
+        "delta_sigma",
+        "deltasigma",
+        "sigma1_minus_sigma3",
+        "q_diff",
+        "deviator_sigma",
+    ),
+    "p'": (
```
indigocoral:
```suggestion
    "Mean Effective Stress": (
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "p'",
+        "pprime",
+        "p_prime",
+        "p",
+        "mean_stress",
+        "meanstress",
+        "mean_effective_stress",
```
indigocoral, line 112: maybe also this:

```suggestion
        "mean_effective_stress",
         "mean effective stress",
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "mean_effective_pressure",
+        "effective_stress_mean",
+        "effective_mean_stress",
+        "p_eff",
+    ),
+    "q": (
```
indigocoral:
```suggestion
    "Deviatoric Stress": (
```

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
...
+                return
+
+            selected_file = Path(csv_path)
+            if selected_file.suffix.lower() != ".csv":
+                messagebox.showerror(
+                    "Import Error",
+                    "The selected file is not a CSV file. "
+                    "Please choose a file with .csv extension.",
+                )
+                return
```
rfaasse: Since we specify the file type in the `askopenfilename` function, I don't think we can ever end up with a 'non-csv' file if I understand it correctly. In that case, I think this check is not necessary and can be removed

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "q",
+        "q_stress",
+        "deviator_stress",
+        "von_mises",
+        "deviatoric_stress",
+        "deviatoricstress",
```
indigocoral, line 126:
```suggestion
        "deviatoricstress",
        "deviatoric stress",
```

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
...
+                test_display_name=current_test_display_name,
+            )
+            if column_mapping is None:
+                return
+
+            self._controller.import_csv_data(
+                selected_file,
+                column_mapping=column_mapping,
+                target_test_type=current_test_internal_name,
+            )
```
rfaasse: As Marjan also mentioned, I think most of this logic can be moved to the result_manager. My expectation of a GUI in this case would be to only provide the model with a file name and the model to handle the rest.

In this case that would mean: the GUI provides a file-name to the model -> the model provides the headers to the GUI -> The GUI displays the pop-up and let's the user choose -> After the choice is made for the mapping, the mapping is provided to the model, such that it can complete the import.

Let me know if you agree!

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    ),
+    "shear_strain_xy": (
```
indigocoral: As this is the tensor shear strain component, I think we should let the user know that this is not the engineering shear strain (gamma_xy). Perhaps in the future we should make sure that we can handle both of them.
```suggestion
    "Tensor Shear Strain": (
```

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
...
+
+            if self.main_frame:
+                self.main_frame.redraw_plots()
+
+        except Exception as e:
+            messagebox.showerror("Import Error", f"Failed to import CSV data.\n\n{e}")
+
+    def _show_csv_header_mapping_popup(
```
rfaasse: Would it be possible to have the pop-up at the center of the app? Right now (at least for me) it pops up in the top left corner.

rfaasse: Just a suggestion, the following makes it slightly more clear in my opinion what the popup is used for:
```suggestion
    def _show_csv_header_mapping_selection_popup(
```

rfaasse: I feel this function is on the long side. Could you break it up into a few smaller ones to improve readability?

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +373,222 @@ def _import_lab_results(self):
...
+                return
+
+            expected_headers = get_expected_columns_for_test_type(
+                current_test_internal_name
+            )
+            suggested_mapping = suggest_csv_column_mapping(
+                file_headers, expected_headers
+            )
+
+            column_mapping = self._show_csv_header_mapping_popup(
```
rfaasse: I really like the mapping functionality. I think it's the only way to provide the user with enough options for importing the data (we'll never know all header columns that will exist)

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -144,13 +147,13 @@ def test_import_lab_results_dict_raises_for_invalid_input(self):
         result_manager = ResultManager(current_test_getter)
 
         with self.assertRaises(ValueError):
-            result_manager.import_lab_results_dict(None)
+            result_manager.import_lab_results_dict(cast(Any, None))
```
rfaasse, line 150: Just for my understanding: why do we need the cast?

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+                "Triaxial,0.0,0.0,,\n"
+                "Triaxial,-0.1,120.0,,\n"
+                "Direct Simple Shear,,,0.0,0.0\n"
+                "Direct Simple Shear,,,0.05,40.0\n",
+                encoding="utf-8",
+            )
+
+            result_manager.import_csv_lab_results(csv_file)
+
+        current_test_type = TRIAXIAL
```
rfaasse: This seems duplicate (since we already set this in 239 and don't seem to change it in between)

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+                "sigma_1": [-100.0, -250.0],
+                "sigma_3": [-100.0, -100.0],
+                "vol_strain": [0.0, -0.01],
+                "p'": [-100.0, -150.0],
+                "q": [0.0, 150.0],
+                "sigma1_sigma3_diff": [0.0, 150.0],
+            },
+        )
+
+    def test_import_csv_lab_results_with_test_type_column_routes_data_per_test(self):
```
rfaasse, line 269: I don't think we need the 'test_type' header to be taken into account, since we require one file per test type anyway

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+        current_test_type = DIRECT_SHEAR
+        self.assertDictEqual(
+            result_manager.get_experimental_results(),
+            {
+                "shear_strain_xy": [0.0, 0.05],
+                "shear_stress_xy": [0.0, 40.0],
+            },
+        )
+
+    def test_import_csv_lab_results_supports_cp1252_encoding(self):
```
rfaasse, line 343: Interesting, is this encoding used often? And are there others which are used commonly?

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+                "sigma_1": [-100.0, -250.0],
+                "sigma_3": [-100.0, -100.0],
+                "vol_strain": [0.0, -0.01],
+                "p'": [-100.0, -150.0],
+                "q": [0.0, 150.0],
+                "sigma1_sigma3_diff": [0.0, 150.0],
+            },
+        )
+
+    def test_import_csv_lab_results_supports_common_lab_header_labels(self):
```
rfaasse, line 409: Nice that we have these defaults!

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+                    "yy_strain": "epsilon",
+                },
+            )
+
+        self.assertDictEqual(
+            result_manager.get_experimental_results(),
+            {
+                "q": [0.0, 120.0],
+                "yy_strain": [0.0, -0.1],
+                "sigma1_sigma3_diff": [0.0, 120.0],
```
rfaasse: I don't understand how the sigma1_sigma3_diff is calculated here, since there is only q and yy_strain, so I'm probably missing some background here. Could you elaborate?

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+        current_test_type = CRS
+        self.assertDictEqual(
+            result_manager.get_experimental_results(),
+            {},
+        )
+
+        current_test_type = DIRECT_SHEAR
+        self.assertDictEqual(result_manager.get_experimental_results(), {})
+
+    def test_import_csv_lab_results_keeps_manual_mapping_for_multiple_test_types(self):
```
rfaasse: If I understand the test correctly, it's more about the manual mapping in combination with only importing the data related to that specific test type (in this case triaxial). Maybe you could rename it to reflect this a bit more clearly?

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,401 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+
+        # CRS should not have any data since it wasn't the active test
+        current_test_type = CRS
+        self.assertDictEqual(
+            result_manager.get_experimental_results(),
+            {},
+        )
+
+        current_test_type = DIRECT_SHEAR
+        self.assertDictEqual(result_manager.get_experimental_results(), {})
```
rfaasse, line 545: Curious that the assert is formatted on one or multiple lines, while the line is identical. I see black formatting just keeps it like this, so no changes required, but it's strange to me

kratos_element_test/model/test/test_result_manager.py
rfaasse, line 1: Nice set of tests for the csv import!

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    "time_steps": (
+        "time_steps",
+        "time",
+        "time_h",
+        "time_hours",
+        "time_hour",
+        "time_s",
+        "time_sec",
+        "time_seconds",
+        "hours",
```
rfaasse, line 172: I'm not too sure about these ones, since time_steps are integers. If we'd accept e.g. time_hours, to me that implies that we convert to steps somehow. I'd propose to remove these defaults

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+def _parse_float(value: str, line_number: int, column_name: str) -> Optional[float]:
+    stripped = (value or "").strip()
+    if not stripped:
+        return None
+
+    compact = stripped.replace(" ", "")
+    if "," in compact and "." not in compact:
+        compact = compact.replace(",", ".")
+    elif "," in compact and "." in compact:
+        compact = compact.replace(",", "")
```
rfaasse, line 302: I don't think this part is covered in the unit tests (e.g. numbers like 12,34 or 12,000.0)

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+
+    class CustomDialect(csv.Dialect):
+        delimiter = best_delimiter
+        quotechar = '"'
+        doublequote = True
+        skipinitialspace = False
+        lineterminator = "\r\n"
+        quoting = csv.QUOTE_MINIMAL
+
+    return CustomDialect
```
rfaasse, line 371: Is it realistic that the csv.Sniffer().sniff(sample, delimiters=",;\t |") line doesn't return anything (at least I see this part isn't really covered in the unit tests)? And in that case should we try to recover like this, or would it be okay to just not accept the file? If that's the case we could simplify by removing this more custom dialect finding.

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+            continue
+
+        return decoded
+
+    details = (
+        "; ".join(decode_errors[:3]) if decode_errors else "unknown decoding error"
+    )
+    raise ValueError(
+        f"Could not decode CSV file '{csv_file}'. "
+        "Supported encodings include UTF-8, UTF-16 and Windows-1252. "
```
rfaasse: Should the 'latin-1' be added to this message as well?

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        if expected and canonical_key not in expected:
+            continue
+        if canonical_key in mapping:
+            continue
+        mapping[canonical_key] = header
+
+    return mapping
+
+
+def parse_csv_lab_results(
```
rfaasse: This function is in my opinion very lengthy. Could you split it up into a few smaller ones, such that it's easier to read?

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+    csv_stream = io.StringIO(csv_text, newline="")
+
+    sample = csv_stream.read(2048)
+    csv_stream.seek(0)
+
+    dialect = _detect_csv_dialect(sample)
+
+    reader = csv.DictReader(csv_stream, dialect=dialect)
+    if not reader.fieldnames:
+        raise ValueError("CSV file has no header row")
```
rfaasse: When I import a csv without headers, I seem to just get the first row of numbers as headers, so I'm not sure this check is working as expected

rfaasse (commented):
Hi Gennady, thank you for implementing this nice functionality! I tested it and it works well and is robust as far as I could see! Functionality wise, I have only one real suggestion for a change, which is about the default test (similar to Marjans comments). Next to that, I have some questions and suggestions about the code itself.

kratos_element_test/plotters/lab_result_overlay_registry.py
indigocoral: I would like to keep "Lab results" as this was what we agreed on before.

kratos_element_test/model/io/lab_results_csv_parser.py
indigocoral: In response to this, I think we have to keep time_hours (as this is what CRS test gets as inputs: time in hours) and then remove these ones:
"time_s",
"time_sec",
"time_seconds",

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "q_stress",
+        "deviator_stress",
+        "von_mises",
+        "deviatoric_stress",
+        "deviatoricstress",
+    ),
+    "shear_strain_xy": (
+        "shear_strain_xy",
+        "strain_xy",
+        "gamma_xy",
```
indigocoral, line 133:
```suggestion
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "epsilon_xy",
+        "exy",
+    ),
+    "shear_stress_xy": (
+        "shear_stress_xy",
+        "shear_xy",
+        "gamma_tau",
```
indigocoral: gamma is not standard for shear stress. tau is used for shear stress.

```suggestion
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "deviatoricstress",
+    ),
+    "shear_strain_xy": (
+        "shear_strain_xy",
+        "strain_xy",
+        "gamma_xy",
+        "epsilon_xy",
+        "exy",
+    ),
+    "shear_stress_xy": (
```
indigocoral:
```suggestion
    "Shear Stress": (
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "tau_xy",
+        "shear_stress",
+    ),
+    "sigma_yy": (
+        "sigma_yy",
+        "sigmayy",
+        "sigma_prime_yy",
+        "vertical_effective_stress",
+        "vertical_stress",
+        "effective_vertical_stress",
```
indigocoral, line 151:
```suggestion
        "effective_vertical_stress",
        "effective vertical stress",
        "vertical effective stress",
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "exy",
+    ),
+    "shear_stress_xy": (
+        "shear_stress_xy",
+        "shear_xy",
+        "gamma_tau",
+        "tau_xy",
+        "shear_stress",
+    ),
+    "sigma_yy": (
```
indigocoral:
```suggestion
    "Vertical Effective Stress": (
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "sigma_yy",
+        "sigmayy",
+        "sigma_prime_yy",
+        "vertical_effective_stress",
+        "vertical_stress",
+        "effective_vertical_stress",
+    ),
+    "sigma_xx": (
```
indigocoral:
```suggestion
    "Horizontal Effective Stress": (
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "sigma_xx",
+        "sigmaxx",
+        "sigma_prime_xx",
+        "horizontal_effective_stress",
+        "horizontal_stress",
+        "effective_horizontal_stress",
+    ),
```
indigocoral, line 164:
```suggestion
    "effective horizontal stress",
    "horizontal effective stress",
    ),
```

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+        "time_sec",
+        "time_seconds",
+        "hours",
+    ),
+}
+
+_EXPECTED_COLUMNS_BY_TEST: Dict[str, List[str]] = {
+    "triaxial": [
+        "yy_strain",
+        "sigma1_sigma3_diff",
```
indigocoral: As mentioned, maybe we can remove this as it will be calculated by sigma_1 and sigma_3 if I'm correct.

kratos_element_test/model/io/lab_results_csv_parser.py
indigocoral: Yes.. I also just removed the title and it shows me the blank space there...
Maybe it's better to add a check that if the column doesn't have any header (i.e. either it doesn't have any header at all or the header is completely empty) then it shows the column number maybe? Then the user can choose the column number in the mapping.

kratos_element_test/model/io/lab_results_csv_parser.py
```diff
@@ -0,0 +1,592 @@
...
+            selected_source = (source_column or "").strip()
+            if not selected_source:
+                continue
+
+            if selected_source not in header_set:
+                raise ValueError(
+                    f"Mapped source column '{selected_source}' was not found in CSV headers"
+                )
+
+            # Replace any existing source mapped to this canonical key.
```
indigocoral: I've seen some of these comments also in other places. If they are not needed, please remove them.

indigocoral (commented):
Once again, thank you, Gennady, for all the hard work you did on getting the CSV import in. I know these are a lot of comments, but many of them are just very tiny removals or additions to the mapping labels.
I hope they are clear! 
Please also update the verison in both places (ui_constants.py and also the pyproject.toml) to "Version 0.6.1"

kratos_element_test/plotters/lab_result_overlay_registry.py
markelov208: reverted. I changed it because `Experimental `is used in other locations. Right now, I've changed all `Experimental`.

kratos_element_test/view/ui_menu.py
markelov208: moved lines 402-428 to `result_manager.py`

kratos_element_test/view/ui_menu.py
markelov208: reverted default Triaxial. I prefer the user chooses a test by her/himself.

main.py
markelov208: As Richard suggested it is remained as main.py. It works as startElementTest.exe. It was created by AI to run/debug ElementTest from VS code.

kratos_element_test/view/kratos_element_test_gui.py
markelov208: Actually AI has added these lines to ensure that the script can be run directly (not just as a module)'. I tried this script with/without these lines and it runs or not depending on the environment. So I do not see an effect of these lines. Perhaps, @rfaasse can help to understand them?

kratos_element_test/view/soil_test_input_view.py
markelov208: removed

kratos_element_test/view/ui_builder.py
markelov208: Good catch, no, removed.

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: corrected

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: removed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: removed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: done

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: done

kratos_element_test/controller/element_test_controller.py
markelov208: moved to the previous location

main.py
markelov208: renamed; however, Marjan has suggested to keep only one file in the view folder.

kratos_element_test/controller/element_test_controller.py
markelov208: removed

kratos_element_test/view/ui_menu.py
markelov208: Yes, it may happen when the user can go directly to the import menu.

kratos_element_test/view/soil_test_input_view.py
markelov208: We can have the triaxial test as default. I think it is better if the user chooses the test firstly. Reverted.

kratos_element_test/view/ui_builder.py
markelov208: removed

kratos_element_test/model/test/test_soil_test_input_manager.py
markelov208: reverted changes

kratos_element_test/model/main_model.py
markelov208: AI moved it here and I thought that it was better to see the import where it was used. Now it is reverted.

kratos_element_test/model/material_input_manager.py
markelov208: reverted

kratos_element_test/view/ui_menu.py
markelov208: This function does not block to open any file. It shows files with csv extension but the user can type any name.

kratos_element_test/model/result_manager.py
markelov208: Thank you for the suggestion. It is cleaner as well we do not have the return back.

kratos_element_test/view/ui_menu.py
markelov208: changed. Hope now it matches your ideas.

kratos_element_test/view/ui_menu.py
markelov208: As far as I understand `messagebox `does not allow one to set a position. I've checked other calls of it in this file and they are the same. In my case, the popup windows in the center unless I move the main window to another screen ;)

markelov208: changed

markelov208: Indeed, refactored

kratos_element_test/model/test/test_result_manager.py
markelov208: AI says it is for type checker to suppresses the warning.

kratos_element_test/model/test/test_result_manager.py
markelov208: removed

kratos_element_test/model/test/test_result_manager.py
markelov208: As far as I understand `q=sigma_1-sigma_3`

kratos_element_test/model/test/test_result_manager.py
markelov208: As far as I understand cp1252 can be used in Spain. ;) Based on AI, utf-8-sig is very common when utf-16, utf-16-le, utf-16-be are rare.

kratos_element_test/model/test/test_result_manager.py
markelov208: changed to `test_import_csv_lab_results_with_manual_mapping_imports_only_to_active_test_type`

kratos_element_test/model/test/test_result_manager.py
markelov208: The starting point was Marjan's example python script, which has data for all three tests. Then it has been evolved to use a CSV file import for a one test but the functionality is still in `lab_results_csv_parser.py`.

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: removed time in seconds

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: removed "p"

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added  "mean effective stress",

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added         "deviatoric stress",

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: kept because it is used, for example, in MatplotlibPlotter

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: removed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: changed

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: if `sigma_1 `and `sigma_3 `are not available but the user has the difference? Just in case of such a user ;)

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: removed obvious comments

kratos_element_test/model/main_model.py
markelov208: removed `return`

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: Included

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: refactored

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added an error when only numbers are in the first line and two tests for US and European formats.

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: added unit tests.

kratos_element_test/model/io/lab_results_csv_parser.py
markelov208: As far as I understand this comment is related to a CSV file without headers. Now this is treated separately. Just in case a few unit tests are added for` _detect_dialect` function.

kratos_element_test/view/ui_menu.py
markelov208: moved behind the imports

kratos_element_test/model/result_manager.py
markelov208: Currently, in the UI chain `target_test_type` is equal to `current_test`.  However, it give a certain flexibility for the import, for example, the import is not dependent on the UI.

markelov208 (commented):
Hi Marjan and Richard, thank you very much for this hard work. Hope I answered all your comments properly.

kratos_element_test/view/kratos_element_test_gui.py
markelov208: AI:  in your current environment/run mode, it may be redundant. It mainly protects portability across different launch methods.

kratos_element_test/view/ui_menu.py
rfaasse: Clear thanks! I did find one more issue regarding drawing, but whether that's relevant depends on if we keep the default test or not.

kratos_element_test/view/ui_menu.py
rfaasse: I had the same comment somewhere else. I think we should discuss this. Right now, on main, the behavior is such that triaxial is the default. I don't see a reason why this PR about the CSV import should change that behavior. If it is not necessary to change it and if the behavior change is indeed unrelated to the import (which I think it is), in my opinion we should not change it (so keep the default triaxial test).

main.py
rfaasse: Ah okay, in that case, we can remove this main.py (since at this moment it's not used, we can just run the one in view) 👍

main.py
rfaasse: I think you can also run it locally by running the kratos_element_test/view/kratos_element_test_gui.py file (it has the same main)

kratos_element_test/model/test/test_result_manager.py
rfaasse: Okay, since it's in the test, I think that could make sense (I'm not sure about casting  to an Any, but I also don't know enough about python to have a strong opinion about this)

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,431 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+                "Vertical Strain": [0.0, -0.1],
+                "sigma_1 - sigma_3": [0.0, 120.0],
+            },
+        )
+
+        current_test_type[0] = DIRECT_SHEAR
+        self.assertDictEqual(
+            result_manager.get_experimental_results(),
+            {},
+        )
```
rfaasse: I think this can be simplified slightly (the current test type doens't need to be a list):

```suggestion
        current_test_type = TRIAXIAL
        current_test_getter = lambda: current_test_type
        result_manager = ResultManager(current_test_getter)

        with TemporaryDirectory() as tmp_dir:
            csv_file = Path(tmp_dir) / "multi_test_lab.csv"
            csv_file.write_text(
                "test_type,yy_strain,sigma1_sigma3_diff,shear_strain_xy,shear_stress_xy\n"
                "Triaxial,0.0,0.0,,\n"
                "Triaxial,-0.1,120.0,,\n"
                "Direct Simple Shear,,,0.0,0.0\n"
                "Direct Simple Shear,,,0.05,40.0\n",
                encoding="utf-8",
            )

            result_manager.import_csv_lab_results(csv_file)

        current_test_type = TRIAXIAL
        self.assertDictEqual(
            result_manager.get_experimental_results(),
            {
                "Vertical Strain": [0.0, -0.1],
                "sigma_1 - sigma_3": [0.0, 120.0],
            },
        )

        current_test_type = DIRECT_SHEAR
        self.assertDictEqual(
            result_manager.get_experimental_results(),
            {},
        )
```

kratos_element_test/model/test/test_result_manager.py
```diff
@@ -183,6 +186,431 @@ def test_import_experimental_results_maps_type_name_to_storage_key(self):
...
+                "Triaxial,0.0,0.0,,\n"
+                "Triaxial,-0.1,120.0,,\n"
+                "Direct Simple Shear,,,0.0,0.0\n"
+                "Direct Simple Shear,,,0.05,40.0\n",
+                encoding="utf-8",
+            )
+
+            result_manager.import_csv_lab_results(csv_file)
+
+        current_test_type[0] = TRIAXIAL
```
rfaasse: I think this assignment is not necessary, since it's already set to TRIAXIAL at this point

kratos_element_test/model/io/lab_results_csv_parser.py
rfaasse: My question was more about if it's realistic that we ever get to line 346 (i.e. if it ever happens that `dialect = csv.Sniffer().sniff(sample, delimiters=",;\t |")` does not detect it correctly. If it never happens realistically, we wouldn't need the extra code. But no action needed here 👍

kratos_element_test/view/ui_menu.py
```diff
@@ -334,13 +368,175 @@ def _import_lab_results(self):
...
+                title="Select CSV data file",
+                filetypes=[("CSV files", "*.csv")],
+            )
+            if not csv_path:
+                return
+
+            result = self._controller.prepare_csv_import(
+                csv_path, current_test_display_name
+            )
+            if result is None:
```
rfaasse: I think this could be clarified a bit with a rename
```suggestion
            csv_import_info = self._controller.prepare_csv_import(
                csv_path, current_test_display_name
            )
            if csv_import_info is None:
```

kratos_element_test/view/ui_menu.py
rfaasse: Yes, this was what I had in mind, thank you!

kratos_element_test/view/ui_menu.py
rfaasse: Okay, thanks for looking into it, then we'll keep it as is!

kratos_element_test/model/test/test_result_manager.py
rfaasse: I think there is still something not completely correct in this test, I have created separate comments for it for clarity

kratos_element_test/model/io/lab_results_csv_parser.py
rfaasse: Nice, thanks, this is a lot clearer!

kratos_element_test/view/ui_menu.py
```diff
@@ -511,83 +470,72 @@ def _show_csv_header_mapping_popup(
...
 
-        if not is_confirmed:
-            return None
-        return result
+        if is_confirmed.get():
+            return {
+                k: v.get() for k, v in vars_by_expected.items() if v.get() != "<skip>"
+            }
+
+        return suggested_mapping
```
rfaasse: Shoudn't we return None here? That way, if we cancel the pop-up, there won't be any import done (if I understand the flow correctly)

rfaasse (commented):
Hi Gennady, thanks a lot for all the hard work on this PR! I re-reviewed and re-tested the latest changes and I only have a few very minor comments left. I think this is very close to being merged!

indigocoral (commented):
Hi Gennady, thank you very much for all the hard work you did on this PR. I tried testing it, and it works. Nice that now the Triaxial test is the default. Also thanks for applying my review comments.
The only thing remaining is changing the version. Please update the version in both places (ui_constants.py and also the pyproject.toml) to "Version 0.6.1"

main.py
markelov208: removed

main.py
markelov208: removed

kratos_element_test/model/test/test_result_manager.py
markelov208: simplified

kratos_element_test/model/test/test_result_manager.py
markelov208: removed

kratos_element_test/view/ui_menu.py
markelov208: changed

kratos_element_test/view/ui_menu.py
markelov208: Indeed, it is clearer. Done

markelov208:
> Hi Gennady, thank you very much for all the hard work you did on this PR. I tried testing it, and it works. Nice that now the Triaxial test is the default. Also thanks for applying my review comments. The only thing remaining is changing the version. Please update the version in both places (ui_constants.py and also the pyproject.toml) to "Version 0.6.1"

Hi Marjan, updated. Thank you for checking. I think it was also a hard work.

markelov208:
> Hi Gennady, thanks a lot for all the hard work on this PR! I re-reviewed and re-tested the latest changes and I only have a few very minor comments left. I think this is very close to being merged!

Hi Richard, many thanks for the valuable suggestions.

rfaasse (approved):
Thank you for all the hard work on this! To me it is ready to go!

